PR #4, opened 2026-05-06
Fix MiniMax TTS playback on device

icodeface:
## Summary
- fix MiniMax TTS playback by using non-streaming API (`stream: false`)
- decode complete MP3 response with ffmpeg and play via persistent reader thread
- avoid streaming API issues with cumulative snapshots and MP3 delta extraction

## Test plan
- [x] User verified on target device that TTS playback works without duplication
- [x] `/Users/apple/IdeaProjects/aiden-hardware-demo/build-host/tests/aiden_tests --test-case="*stream parser*"`

## Notes
Initial attempts to preserve streaming with MP3 delta extraction caused audio loss on device. Non-streaming approach successfully delivers audio without duplication or loss.


icodeface:
https://platform.minimax.io/docs/api-reference/speech-t2a-http

PR #65, opened 2026-05-28
Merge/config test button

codesister:
## Summary by CodeRabbit

* **New Features**
  * Added a per-section "测试" (Test) action to configuration pages to run self-checks.
  * Coverage includes proxy, model, TTS, STT, audio, HID, agent settings, and search provider checks.
  * Shows aggregated per-check pass/fail details in a dismissible toast with clear success/failure state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/AidenAI-IO/aiden-hardware-demo/pull/65?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


HuangHan96:
@codesister 需要关闭所有的 Conversation

aiden-lightning:
## Code Review Summary

**Verdict: Changes requested**

### ⚠️ Blocking / should fix before merge
- `src/config_web.cpp` builds the model API-key validation request by putting the raw API key into a `curl` command-line argument (`Authorization: Bearer ...`). On Linux, command-line args are observable via `/proc`/`ps` while the request is running, so clicking “测试” can leak the user’s model API key to other local processes. Please avoid passing secrets in argv; write headers/body to a private temp file or use a curl config file/stdin with restrictive permissions, then delete it.
- The endpoint validation also ignores the proxy values being tested/saved. `build_curl_proxy_arg()` is added but never used, so model/TTS/STT endpoint checks will fail in environments that require `http_proxy`/`https_proxy`/`all_proxy` even when the proxy section is correctly configured. Please apply the selected proxy config to all outbound `curl` checks, or make the UI clear that these checks intentionally bypass proxy settings.

### Notes
- I like the UI direction: per-section test buttons and structured pass/fail output are useful for device setup.
- The implementation uses shell quoting consistently for user-controlled values; that helps reduce command injection risk, but secret-in-argv is still a separate exposure.

### Verification
- Reviewed the diff for `src/config_web.cpp` and `src/config_web_html.h` against `origin/main`.
- Did not run a firmware build locally; review is based on static diff inspection.

HuangHan96 (approved):

PR #83, opened 2026-06-01
feat: add device-wide system environment

icodeface:
## Summary
- Add /userdata/system/env as a device-wide shell environment loaded by services and login shells
- Move proxy configuration out of agent.toml and into system environment variables
- Add config_web system env editing plus proxy validation for malformed values with invisible whitespace or embedded proxy assignments
- Make the quick memory benchmark target configurable through environment variables

## Tests
- go test ./... -count=1
- go test ./internal/agent -count=1
- python3 -m py_compile benchmark/scripts/quick_memory_test.py
- cmake --build build-host --target aiden_tests
- ctest --output-on-failure
- sh scripts/test_system_env.sh
- sh scripts/test_ota_init.sh
- sh scripts/test_build_scripts.sh
- sh -n overlay/oem/usr/bin/aiden-env-run overlay/etc/profile.d/aiden-env.sh overlay/etc/init.d/S52frame_service overlay/etc/init.d/S53audio_service overlay/etc/init.d/S53agent overlay/etc/init.d/S54ota overlay/etc/init.d/S56config_web scripts/test_system_env.sh scripts/test_ota_init.sh
- manual config_web API runtime check

## Summary by CodeRabbit

* **New Features**
  * Central device system env file (/userdata/system/env) for proxy and environment variables.
  * Config Web UI: editor to view/save system env and apply proxy settings.
  * Services and SSH logins now load environment via a launcher wrapper so system env is applied at startup.
  * Benchmarks/tests and tooling honor environment overrides for agent/SSH endpoints.

* **Documentation**
  * Docs updated to describe centralized system env usage and migration from per-service proxy settings.


src/agent/internal/netproxy/proxy.go
```diff
@@ -0,0 +1,174 @@
...
+		return firstEnv("ALL_PROXY", "all_proxy")
+	}
+}
+
+func ProxyFromEnvironment(req *http.Request, allowedSchemes ...string) (*url.URL, error) {
+	name, raw := EnvForRequest(req)
+	if raw == "" {
+		return nil, nil
+	}
+	if req != nil && req.URL != nil && Bypass(req.URL.Hostname(), req.URL.Port(), NoProxyFromEnvironment()) {
```
aiden-lightning: This no longer preserves the standard `http.ProxyFromEnvironment` loopback bypass. With `HTTP_PROXY` or `ALL_PROXY` set and no `NO_PROXY`, requests to `127.0.0.1` are proxied instead of going direct. I hit this locally: `go test ./internal/ota` failed against `httptest` servers with `net/http: HTTP/1.x transport connection broken: unexpected EOF`, while clearing proxy env lets the same download test pass. Please keep the localhost/loopback bypass in `ProxyFromEnvironment` or inject the default no_proxy before evaluating it, and cover it with a test.

src/config_web.cpp
```diff
@@ -1250,6 +1315,252 @@ std::string read_file_contents(const char* path, size_t max_size = 64 * 1024) {
...
+                                      : trim_trailing_newlines(result.output);
+            if (error->empty()) {
+                *error = "system env syntax check failed";
+            }
+        }
+        return false;
+    }
+
+    SystemProxy proxy;
+    if (load_system_env_proxy(tmp_template, &proxy) &&
```
aiden-lightning: `validate_system_env_content` ends up sourcing the just-submitted file via `load_system_env_proxy`. That makes validation execute arbitrary shell from the web request before the file is saved, so rejected content can still have side effects. It also ignores load failures, so content such as `touch /tmp/pwn; exit 0` passes `sh -n`, executes during validation, is saved, and then makes every `aiden-env-run` service exit before `exec`. The config endpoint should parse allowed assignment lines without sourcing user input, and treat unsupported/control-flow content as a validation error.

src/config_web.cpp
```diff
@@ -1939,11 +2295,47 @@ load();
...
+    }
+    return false;
+}
+
+std::string validate_proxy_url(const std::string& url) {
+    if (url.empty()) return "";
+    if (has_proxy_whitespace(url)) return "proxy URL contains whitespace; use one export assignment per line";
+    if (has_embedded_proxy_assignment(url)) return "proxy URL contains another proxy assignment; use one export assignment per line";
+    if (has_duplicate_proxy_scheme(url)) return "duplicate scheme in proxy URL";
     return starts_with(url, "http://") || starts_with(url, "https://") ||
```
aiden-lightning: The web-side validator is looser than the Go runtime. It accepts any string that starts with an allowed prefix, so `http://` passes despite `netproxy.Parse` rejecting it as missing a host, and it advertises/accepts `socks4://` and `socks4a://` while the agent and OTA runtime only allow `http`, `https`, and `socks5`. Saving one of these values through the UI writes `/userdata/system/env`, restarts the agent, and then `ProxyConfigFromEnvironment().Validate()` fails at startup. Please align this with `netproxy.Parse`: absolute URL with a host and the same allowed schemes.

aiden-lightning (changes requested):
I found a few issues that should be fixed before merging. The main theme is that the new environment/proxy path is not consistently validated and can regress local HTTP traffic when proxy variables are present.

src/config_web.cpp
```diff
@@ -1250,6 +1299,218 @@ std::string read_file_contents(const char* path, size_t max_size = 64 * 1024) {
...
+    if (!validate_system_proxy_values(proxy, error)) {
+        return false;
+    }
+
+    std::string existing = read_file_contents(path.c_str(), kMaxSystemEnvSize);
+    std::ostringstream out;
+    std::istringstream input(existing);
+    std::string line;
+    while (std::getline(input, line)) {
+        if (is_system_proxy_assignment_line(line)) {
```
aiden-lightning: This skips the entire line whenever the first assignment is a proxy key, which can delete unrelated environment variables. The new parser supports multiple assignments on one line, and users may reasonably have something like `export HTTP_PROXY=http://proxy:7890 OPENAI_API_KEY=...` in `/userdata/system/env`. Saving the proxy section through `/api/config` would drop `OPENAI_API_KEY` because the whole line is omitted before the managed proxy block is appended. Please either parse existing content and remove only proxy-key assignments, or reject/normalize multi-assignment lines so unrelated values cannot be lost.

src/system_env_parser.cpp
```diff
@@ -0,0 +1,352 @@
...
+    }
+    if (has_embedded_proxy_assignment(trimmed)) {
+        return "proxy URL contains another proxy assignment; use one export assignment per line";
+    }
+    if (has_duplicate_proxy_scheme(trimmed)) {
+        return "duplicate scheme in proxy URL";
+    }
+
+    size_t scheme_end = trimmed.find("://");
+    if (scheme_end == std::string::npos || scheme_end == 0 || !proxy_host_present(trimmed)) {
```
aiden-lightning: This still does not fully match the Go runtime's `netproxy.Parse`/`net/url.Parse` validation. The C++ check only verifies that some host-looking text exists, so malformed authorities such as `http://%zz` pass here but are rejected by Go URL parsing at agent startup. That means the UI can still save a proxy value that restarts the agent into a validation failure. Please add malformed-authority tests and make the C++ validator reject the same class of invalid URLs as the runtime.

aiden-lightning (changes requested):
Thanks for addressing the previous review. The original three issues are mostly covered now, but I found two remaining correctness issues that can still break persisted environment configuration or allow the UI to save proxy values the Go runtime rejects.

src/system_env_parser.cpp
```diff
@@ -0,0 +1,489 @@
...
+    if (has_duplicate_proxy_scheme(trimmed)) {
+        return "duplicate scheme in proxy URL";
+    }
+    std::string bad_escape = invalid_percent_escape(trimmed);
+    if (!bad_escape.empty()) {
+        return "invalid URL escape \"" + bad_escape + "\"";
+    }
+
+    size_t scheme_end = trimmed.find("://");
+    if (scheme_end == std::string::npos || scheme_end == 0 || !proxy_host_present(trimmed)) {
```
aiden-lightning: This still accepts URL forms that Go's `net/url.Parse` rejects, so the UI can save a proxy that later prevents the agent from starting. For example, `http://proxy.example:bad`, `http://[::1]zzz`, and a URL containing an ASCII control character all return success here because `proxy_host_present` only checks for non-empty host text, while the Go runtime rejects them with invalid port/control-character errors. Please either share a stricter parser or add equivalent authority/control-character validation and tests for these cases, not just malformed `%` escapes.

aiden-lightning (changes requested):
The mixed-assignment preservation fix looks good, but one runtime-parity issue remains in the C++ proxy URL validator.

src/system_env_parser.cpp
```diff
@@ -0,0 +1,537 @@
...
+            return "invalid proxy URL port";
+        }
+        return "";
+    }
+
+    if (host_port.find(']') != std::string::npos || host_port.find('[') != std::string::npos) {
+        return "invalid proxy URL authority";
+    }
+
+    size_t colon = host_port.find(':');
```
aiden-lightning, line 491: The authority check still does not validate host characters the way Go's `net/url.Parse` does. For example, `http://foo\bar`, `http://foo^bar`, and `http://foo|bar` all pass `validate_system_proxy_url` here, but `net/url.Parse` rejects them as invalid host names, so saving one through the proxy UI will still make `ProxyConfigFromEnvironment().Validate()` fail on agent startup. Please add host-character validation/tests or otherwise delegate to an implementation that matches the Go runtime.

aiden-lightning (changes requested):
The latest fixes cover the invalid-port/control-character cases I called out, but the C++ validator still accepts some host characters that Go rejects, so the UI can still persist a proxy value that breaks agent startup.

src/system_env_parser.cpp
```diff
@@ -0,0 +1,622 @@
...
+    std::string authority = trimmed.substr(
+        host_start,
+        host_end == std::string::npos ? std::string::npos : host_end - host_start);
+    size_t at = authority.rfind('@');
+    std::string host_port = at == std::string::npos ? authority : authority.substr(at + 1);
+    if (host_port.empty()) {
+        return "expected absolute proxy URL, for example http://127.0.0.1:7890";
+    }
+
+    if (host_port[0] == '[') {
```
aiden-lightning, line 473: This branch treats any non-empty bracketed host as valid, so values like `http://[zzzz]:7890` and `http://[v1.foo]:7890` pass the config_web validator. Go's `net/url.Parse`, which the agent runtime uses via `netproxy.Parse`, rejects both as invalid host/IP literals, so the UI can still save a proxy value that later prevents agent startup. Please validate bracketed hosts as IPv6/IP literals (or otherwise mirror Go's URL parser) and add regression tests for invalid bracketed hosts.

aiden-lightning (dismissed):
The latest changes fixed the host-character cases from the previous review, but there is still one URL-validation parity gap for bracketed hosts.

aiden-lightning (approved):

PR #94, opened 2026-06-02
feat(agent): bind HTTP server to 0.0.0.0 for LAN access

codesister:
The agent HTTP server was only reachable via usb0 (192.168.42.1) because Go's `:8080` notation bound to IPv6-only on this kernel. Explicitly bind to 0.0.0.0:8080 so the agent Web UI and API are accessible from wlan0 (LAN) as well.

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent now binds to all network interfaces by default, allowing access from other machines.
  * Web UI startup message now displays the localhost URL in a clearer format.

### DIFF
--- a/src/agent/cmd/daemon/main.go
+++ b/src/agent/cmd/daemon/main.go
@@ -77,8 +77,6 @@ func main() {
 	fmt.Printf("📂 Config directory: %s\n", *configDir)
 	if _, port, err := net.SplitHostPort(*addr); err == nil && port != "" {
 		fmt.Printf("🌐 Web UI: http://localhost:%s\n", port)
-	} else {
-		fmt.Printf("🌐 Web UI: http://localhost%s\n", *addr)
 	}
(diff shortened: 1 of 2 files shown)


src/agent/cmd/daemon/main.go
```diff
@@ -74,7 +75,11 @@ func main() {
 
 	fmt.Printf("🚀 Aiden Agent daemon starting on %s\n", *addr)
 	fmt.Printf("📂 Config directory: %s\n", *configDir)
-	fmt.Printf("🌐 Web UI: http://localhost%s\n", *addr)
+	if _, port, err := net.SplitHostPort(*addr); err == nil && port != "" {
+		fmt.Printf("🌐 Web UI: http://localhost:%s\n", port)
+	} else {
+		fmt.Printf("🌐 Web UI: http://localhost%s\n", *addr)
+	}
```
HuangHan96, line 80: 这个 else 是做什么的

PR #173, opened 2026-06-10
feat(agent): migrate benchmark + user_files to agent (8080) + screenshot task recorder

ljunb:
Move all benchmark + user_files HTTP routes from `config_web` (port 80, C++) to the Go agent (port 8080), and add a new perception-task recorder that lets users append screenshot tasks to `perception_v1` from the browser.

## Why

- `config_web` was 4000+ lines of C++ growing more benchmark + LLM logic on top of static config UI — wrong scope for the C++ service.
- `benchmark`/`user_files` are runtime observability, naturally belong with the agent.
- `scripts/files_proxy.py` was a stale dev helper duplicating routes already served by `config_web`.
- Want a way to author perception tasks in-browser instead of hand-editing JSON + dropping screenshots into the suite directory.

## Port topology

| Port | Before | After |
|------|--------|-------|
| 80 | config_web + benchmark + user_files | config_web only (device config) |
| 8080 | agent chat/tools | agent + benchmark + user_files + recorder |
| 8090 | files_proxy.py (dev only) | _removed_ |

URL paths are unchanged — bookmarks/scripts only need the host:port swap.

## What moved (12 routes)

`GET /benchmark`, `/benchmark/suites`, `/benchmark/runs`, `/benchmark/report/<id>`, `/benchmark/status`, `/benchmark/log`, `POST /benchmark/run`, `/benchmark/suites/import`, `/benchmark/suites/delete`, `/benchmark/suites/generate`, `GET /user_files`, `POST /user_files/regenerate`.

Each Go handler is a 1:1 port of the C++ implementation — same response shapes, same status codes, same self-heal logic on `state.json`. The `/benchmark` index HTML was copied verbatim from `de4cc631^:src/config_web.cpp` so look + behavior match.

## New: perception-task recorder

New page `/benchmark/record` lets users:
1. Grab a live screenshot from the device or paste/upload an image
2. Drag a rectangle on the canvas to mark the target UI element
3. Type the target name + what they want the agent to do (2 fields total — task ID is auto-generated)
4. Hit *Generate*: backend calls the configured LLM with the image + metadata, returns a single-task JSON
5. Hit *Append*: writes the screenshot to `benchmark/suites/perception/screenshots/<task_id>.jpg` and appends the task to `perception_v1.json`

Two new endpoints support it:
- `POST /benchmark/suites/generate-perception` — multimodal LLM call; backend backfills coordinate ranges + slug so the rubric text stays consistent regardless of LLM precision
- `POST /benchmark/suites/append-perception` — atomic append to `perception_v1.json` (per-suite mutex, tmp + rename)

Coordinates are emitted in the **0-1000 normalized space** (`x in [501, 536]`) to match `tools_hid.go` semantics from #164. Rubric text now reads `normalized x in [a, b], y in [c, d] (0-1000 normalized space, where 500 is center)`.

## Sync from device

The recorder writes only to the device. To pull tasks back into the repo:

\`\`\`
scripts/sync_perception_from_device.sh <device_ip>
\`\`\`

Diffs device vs. repo, asks for confirmation, copies `perception_v1.json` + `screenshots/` back, stages in git. Doesn't auto-commit — review first.

## Cleanup

- Deleted `scripts/files_proxy.py`
- Deleted `benchmark/suites/full_smoke.json` (no code references)
- Removed 12 routes + 4 helpers + related constants from `src/config_web.cpp` (~1000 lines)
- Updated `tests/config_web_source_test.cpp` accordingly

## Tested

- 25+ Go unit tests covering each new handler — all pass
- End-to-end on real device (192.168.31.107):
  - All 12 migrated endpoints return expected status + body
  - Recorder full flow: capture → drag → generate → append → screenshot lands at `screenshots/<task_id>.jpg`, task appended to `perception_v1.json`
  - History table populates Suite + totals from `manifest.json`
  - Path traversal rejected on `/benchmark/report/`
- C++ test suite (`make test`) still passes after route deletion

## Known limitations (out of scope)

- Existing `report.html` template doesn't embed `tasks/<id>/history.json` — drawer detail view shows empty prompt/response. Disk artifacts are intact; needs a fix in the upstream Python `runner/html_report.py`.
- Screenshot recorder requires a vision-capable LLM (`agent.toml` `[model]` section). Default `gemini-3.5-flash` works; non-vision models will fail at generate-perception.

## Summary by CodeRabbit

* **New Features**
  * Web-based benchmark management (browse runs, run/import/delete suites, live logs/status, HTML reports) and a screenshot recorder for creating perception tasks.
  * LLM-assisted suite and perception-task generation.
  * User Files report viewing and asynchronous regeneration.
  * Device perception sync script to pull benchmark data from devices.

* **Chores**
  * Moved benchmark and user-files handling into the agent service and added benchmark-related config support.
  * Removed legacy proxy server and obsolete plan/spec documents.


ljunb:
Addressed all three review comments in b2c9dbab:

- **`sync_perception_from_device.sh:47`** — switched to single-quoted trap so `$TMP_DIR` expands at EXIT, not at registration.
- **`user_files.go:52`** — extracted `runUserFilesScript()` and reused it from `handleUserFilesRegenerate` so the background path inherits the 30s timeout. A hung script can no longer leak long-lived child processes.
- **`user_files_test.go:62`** — added an immediate `os.Stat(marker)` assertion right after the handler returns to make the test fail under a synchronous implementation. Bumped the script sleep to 200ms for meaningful headroom.

PR #181, opened 2026-06-11
feat(session): detect voice session boundaries

icodeface:
## Summary
- Add a local session-boundary classifier for voice multi-task turns using time gaps, continuation markers, action intent, app divergence, and recent episode state.
- Rotate a newly closed active session into `memory/session_archive/<closed_session_id>/`, recreate an empty active session, and keep archived sessions out of prompt and recall context.
- Return explicit `source` metadata from `recall_session_chunks` results so active indexed chunks are reported as `active` even if their IDs use a `pending-` prefix.
- Count `pending_chunks_recalled` from explicit `source: "pending"` results instead of inferring pending state from `chunk_id` prefixes.
- Document the recall source contract and the session-boundary telemetry metadata fields.

## Testing
- `go test ./internal/agent -run 'Test(SessionRecallTelemetry|RecallSessionChunksTool|SessionMemoryRecall)'`
- `go test ./internal/agent -run 'Test(SessionMemoryStore|SessionMemory|MaintainFilesystemMemory|FormatSessionSummary)'`
- `git diff --check`

Note: full `go test ./internal/agent` still has an unrelated existing prompt test failure in `TestFunctionAgentSystemMessageIncludesGlobalEnvironmentAndDeviceGuidance` around the expected `type "back"` guidance.

### DIFF
--- a/docs/04-agent/configuration.md
+++ b/docs/04-agent/configuration.md
@@ -153,6 +153,9 @@ Optional. Place `memory/extraction.yaml` under the config directory to control s
 | `context_window` | `32000` | Fallback context window for compaction when the active model is not present in `model_specs`. Runtime normally derives this from `ModelResolver.Spec()`; this value is only used for unknown models. |
 | `compress_at_percent` | `50` | Percentage trigger: compaction starts when `prompt_tokens / context_window >= compress_at_percent%`. |
 | `summary_max_chunks` | `10` | Number of chunk summaries kept in the Recent Chunks section of `summary.md`. Older entries move to the archive and are folded into the Rolling Summary. |
+| `session_boundary_enabled` | `true` | Classify each new user turn as continuing the current session or starting a new one. A `new` boundary archives the current `memory/session/` directory and recreates an empty active session. |
+| `session_boundary_short_gap_seconds` | default boundary config | Gap below which a turn is treated as continuation regardless of lexical signals. |
+| `session_boundary_long_gap_seconds` | default boundary config | Gap above which a turn is treated as a fresh session regardless of lexical signals. |
 | `tag_candidates` | see defaults | Candidate keywords matched when tagging chunk summaries. |
 | `entity_suffixes` | `["App","app","APP"]` | Suffixes recognized during entity extraction. |
 
--- a/docs/04-agent/context-lifecycle.md
+++ b/docs/04-agent/context-lifecycle.md
@@ -130,9 +130,17 @@ memory/session/events.jsonl
 
 When compressed history exists, prompt construction wraps the rendered hot window with boundary markers. Those markers are prompt-only and are never persisted.
 
+When session-boundary detection classifies a user turn as a new session, the
+runtime archives the whole active `memory/session/` directory into
+`memory/session_archive/<closed_session_id>/`, recreates an empty
+`memory/session/events.jsonl`, and clears the in-memory conversation window.
+Archived sessions are preserved as logs only; they are not restored, listed,
+switched back into, injected into prompts, or searched by
+`recall_session_chunks`.
+
 ### Session Compaction
 
-Session compaction is handled by `MemoryManager.maintainFilesystemMemory`. It reads `events.jsonl`, decides whether compaction is needed, writes compacted chunks, updates `summary.md`, and replaces `events.jsonl` with the retained hot window.
+Session compaction is handled by `MemoryManager.maintainFilesystemMemory`. It reads active `events.jsonl`, decides whether compaction is needed, writes compacted chunks, updates active `summary.md`, and replaces `events.jsonl` with the retained hot window.
 
 Compression can be triggered by:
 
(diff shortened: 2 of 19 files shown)


aiden-lightning:
## Code Review Summary

**Verdict: Comment / non-blocking findings**

I reviewed PR #181 locally in a detached worktree and did not modify any files.

### Findings

#### ⚠️ Medium — `pending_chunks_recalled` can over-count after pending files are consumed

`consumePendingSessionEvents` compresses a pending file into an active chunk while preserving the `pending-...` chunk ID:

- `src/agent/internal/agent/memory.go:993` uses `ChunkID: pendingChunkIDFromPath(path)`
- `src/agent/internal/agent/runtime.go:833` / `countPendingRecallResults` counts pending recalls solely via `strings.HasPrefix(result.ChunkID, "pending-")`

After maintenance consumes a pending file, the chunk becomes an active indexed chunk, but future recalls of that normal active chunk will still be counted as “pending” because its ID keeps the `pending-` prefix. This does not break recall itself, but it pollutes the new pending-recall telemetry and makes it impossible to distinguish “recalled from a live pending file” from “recalled an already-compressed active chunk that happened to originate from a pending file”.

Suggested fixes:
- Convert consumed pending sessions to normal `chunk_*` IDs and store source session metadata separately; or
- Add a `status` / `pending` field to recall results and count telemetry from that field instead of the ID prefix.

#### 💡 Test coverage — integration test should assert final pending consumption state

`TestRuntimeRunRotatesSessionOnNewBoundary` verifies that rotation creates a pending file and cleans the active hot window, but it does not assert what happens after `releaseMaintenance` + `WaitMaintenance`.

Given pending consumption is a core path in this PR, consider extending the test to assert:

- no `events.pending-*.jsonl` remains after maintenance drains;
- a chunk/index/summary entry was created for the old session;
- recall returns the old-session evidence from the compressed chunk;
- the pending file is not consumed twice.

This would catch regressions where rotation works but async pending consumption silently stops working.

#### 💡 Low — `runningEpisodeExists` comment/implementation mismatch

`BoundaryEpisodeContext.HasRunning` says it is true for `"running"` or `"active"`, but `runningEpisodeExists` only checks `entry.Status == "running"`. I think the implementation is probably correct (finished `active` episodes should not keep biasing continuation), so the comment should likely be narrowed to `"running"` to avoid future confusion.

### Verification run

From `/tmp/review-aiden-hardware-demo-pr181/src/agent`:

- `go build ./...` ✅ passed
- PR focused test command ✅ passed:
  - `go test -run "TestClassifyTurnBoundary|TestSessionRotation|TestMaintainFilesystemMemorySkipsActiveWriteAfterRotation|TestSessionMemoryRecall|TestMaintainFilesystemMemoryConsumesPendingFiles|TestFormatSessionSummaryWithWindow|TestWithSessionBoundaryEnabled|TestRuntimeRunRotatesSessionOnNewBoundary|TestSessionRecallTelemetryCountsPendingResults|TestRuntimeRunSchedulesMemoryMaintenanceAsync|TestRuntimePersistsMemoryUnderConfigDir|TestConcurrentMaintenancePreservesAppendedEvents|TestAppendAndMaintenanceFileLockSerialization|TestMemoryManagerClearAllRemovesFilesystemMemoryArtifacts|TestMemoryManagerClearSessionPreservesLongTermMemory" ./internal/agent`
- `git diff --check origin/main...HEAD` ✅ passed
- `go test ./internal/agent` ❌ failed in the known/unrelated prompt test:
  - `TestFunctionAgentSystemMessageIncludesGlobalEnvironmentAndDeviceGuidance`: missing `type "back"`

Overall I did not find a blocking correctness/security issue in the main session rotation path, but I recommend fixing the telemetry ambiguity or adding a regression test before relying on the new `pending_chunks_recalled` metric.

---
*Reviewed by Hermes Agent*

aiden-lightning (approved):
## Code Review Summary

**Verdict: Approved**

I reviewed PR #181 locally from a detached worktree at `/tmp/aiden-hardware-demo-pr181` against `origin/main`.

### What I checked
- Session-boundary classifier behavior and defaults.
- Runtime rotation ordering before memory retrieval/prompt construction.
- Session archive isolation from active prompt/recall paths.
- Recall telemetry now counting pending results via explicit `source: "pending"` rather than chunk ID prefixes.
- Concurrency around runtime runs, session rotation, compaction, and file locking.
- Backward compatibility of `recall_session_chunks` input handling after `app_name` removal/ignore behavior.

### Verification run
- `git diff --check origin/main...HEAD` ✅
- `go test ./internal/agent -run 'Test(SessionRecallTelemetry|RecallSessionChunksTool|SessionMemoryRecall|SessionBoundary|RuntimeRunRotatesSession|RuntimeRunKeepsSmallSession|RuntimeRunKeepsNeutralFollowUp|RuntimeRunCanceledWhileQueued|RuntimeMemoryContextIgnoresArchivedSessionSummary)'` ✅
- `go test ./internal/agent -run 'Test(SessionMemoryStore|SessionMemory|MaintainFilesystemMemory|FormatSessionSummary)'` ✅

I also ran the full package test suite:

- `go test ./internal/agent` ❌ due to existing/unrelated `TestFunctionAgentSystemMessageIncludesGlobalEnvironmentAndDeviceGuidance`, where the expected text contains `type "back"` but the actual prompt now says `touch_gesture type back/home`. This PR does not modify the prompt files involved in that failure, so I am not treating it as a PR regression.

### Notes
No blocking issues found. The session rotation and active/archive split are covered by focused tests, and the telemetry change avoids the fragile pending-prefix inference issue.

PR #235, opened 2026-06-17
fix(agent): remove runtime memory window cap

icodeface:
Summary:
- remove the runtime ConversationWindowBuffer cap and use buffer memory for restored active session history
- restore all active session chat messages instead of truncating to the previous 10-turn window
- update coverage to assert early active-session messages remain visible in planner context

Tests:
- go test ./internal/agent

### DIFF
--- a/src/agent/internal/agent/chat_history_memory_test.go
+++ b/src/agent/internal/agent/chat_history_memory_test.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	langtools "github.com/tmc/langchaingo/tools"
 )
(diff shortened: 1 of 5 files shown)


aiden-lightning (approved):
Reviewed by Hermes Agent. The change is small and consistent with the stated goal: runtime memory now uses buffer memory, restored active-session chat messages are no longer truncated before being passed as ConversationHistory, and the regression test covers early active-session messages remaining visible in planner context.\n\nVerified locally in an isolated worktree:\n- git diff --check origin/main...HEAD\n- go test ./internal/agent\n\nI also ran go test ./..., but it fails in internal/ota because this macOS OpenSSL build does not support ED25519 (Algorithm ED25519 not found); that package is outside this PR's touched area.

icodeface:
@coderabbitai review

PR #238, opened 2026-06-17
fix(agent): remove follow-up interpretation prompting

icodeface:
## Summary
- Remove natural-language interpretation text for follow-up classifications from planner context
- Keep the raw follow-up classification as a structured signal only
- Add regression coverage to ensure authority-style interpretation text is not injected

## Tests
- go test ./internal/agent
- go test ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified follow-up classification output by removing redundant interpretation text from session context display.


icodeface:
@coderabbitai review

PR #250, opened 2026-06-18
fix(agent): restrict world-state screenshots to verifier

icodeface:
## Summary
- Keep `WorldState.LatestScreenshot` as verifier-only visual evidence so planner and executor world-state prompts stay stable and do not receive duplicate screenshot attachments.
- Preserve executor access to current screenshot observations through the tool scratchpad while validating visual observations by tool capability, dimensions, format, and image payload.
- Lower the default screenshot prune interval from 25 to 2 across Go defaults, C++ TOML defaults, overlay config, docs, and tests.
- Update context lifecycle and configuration docs to match the new screenshot visibility and pruning behavior.

## Tests
- `git diff --check main...HEAD`
- `cd src/agent && go test -count=1 ./...`
- `cmake --build build-host`
- `ctest --test-dir build-host --output-on-failure`

### DIFF
--- a/docs/04-agent/configuration.md
+++ b/docs/04-agent/configuration.md
@@ -20,7 +20,7 @@ TOML 是当前支持的配置格式；JSON 配置已废弃。
 custom_instruction = ""
 max_iterations = -1
 screenshot_keep_n = 3
-screenshot_prune_interval = 25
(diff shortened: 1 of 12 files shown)


icodeface:
@coderabbitai  review

PR #270, opened 2026-06-21
fix(agent): drop finished-episode signal from session boundary scoring

icodeface:
## Summary

Neutral voice follow-ups after a >5-minute pause were never starting a new session, even though the gap clearly exceeded the short-gap threshold. Investigation on-device confirmed the boundary classifier *did* run and deliberately returned `continue` (reason=`active_episode`) — not a missed decision.

Root cause: the `HasActive` signal (a `TaskEpisode` that finished within the long-gap window) re-encoded the time gap rather than adding independent evidence. An episode's `EndedAt` equals the most recent session event, so in the 5–30min mid-range it was *always* true, contributing a fixed +2 that single-handedly forced `continue`. This silently collapsed the mid-range scoring into a 30-minute cliff.

## Change

- Remove `HasActive` from `BoundaryEpisodeContext` and the scoring path entirely (plus the now-dead `recentEpisodeEndedAt`, `BoundaryReasonActiveEpisode`, and the `activeMaxAge` plumbing).
- Keep `HasRunning` — an in-flight task is a genuine state signal that is independent of the time gap.
- Mid-range boundary decisions now rest only on lexical cues, running-episode state, and session size.

## Behavior

- Mid-range gap + neutral input + only a *finished* episode → **new session** (was: continue).
- Running episode + neutral input → still **continue**.
- Short gap (<5min) / long gap (>30min) shortcuts unchanged.

## Tests

- Integration RED→GREEN: `TestRuntimeRunRotatesNeutralFollowUpAfterFinishedEpisode` seeds a large session (>16 events, to defeat the small-session bias) + finished episode + 7.9min gap and asserts the session rotates.
- Unit: `TestClassifyTurnBoundary_FinishedEpisodeDoesNotForceContinue` locks the new contract and verifies a running episode still flips it to continue.
- `recentEpisodeContext` tests updated for the running-only signal.
- `go build ./...`, `go vet`, full `internal/agent` package suite all green.

### DIFF
--- a/src/agent/internal/agent/runtime.go
+++ b/src/agent/internal/agent/runtime.go
@@ -380,8 +380,8 @@ func NewRuntimeWithDeps(cfg Config, models ModelResolver, memories *MemoryManage
 		rt.memoryPlane = NewFilesystemMemoryPlane(filepath.Join(cfg.ConfigDir, "memory"), LoadMemoryExtractionConfig(cfg.ConfigDir), nil)
 		rt.markInterruptedEpisodesBestEffort()
 	}
-	rt.sessionManager = newMemoryManagerSessionManager(memories, func(now time.Time, activeMaxAge time.Duration) BoundaryEpisodeContext {
-		return recentEpisodeContext(rt.memoryPlane, now, activeMaxAge)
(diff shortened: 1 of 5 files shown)


icodeface:
@coderabbitai review

PR #271, opened 2026-06-21
fix(tts): stop duplicate tool-call speech from residual stream buffer

icodeface:
## Summary

Fixes the bug where the assistant speaks its tool-call preamble twice — e.g. you hear "我查下天气" once, then after a pause "我查下天气 + CodeFace, 上海今天有小雨...".

## Root cause

Confirmed by byte-level log correlation across `agent.log`, `audio_service.log`, and the LLM HTTP stream log (turn at 05:07):

- In `phaseDefault`, a planner turn that returns a **tool call** has final streaming enabled, so its content preamble (`我查下天气。`) is streamed into the minimax `sentenceBuffer`.
- The preamble is 6 runes, below `minChunkRunes=18`, so it stays buffered and is never synthesized on its own.
- The same content is also spoken once via `SpeakToolContent` (playback session 46).
- The next **final-answer** turn streams `CodeFace, 上海今天有小雨...`. The leftover preamble is still in the buffer, so it gets prepended and synthesized together (session 47, 74 bytes = 18 + 56) — the user hears "我查下天气" a second time.

`EnableFinalStreaming`'s `resetStreamWriterState` only reset the writer chain, not the TTS adapter's internal sentence buffer.

## Fix

- Add a `ResetBuffer` hook that propagates through the streaming writer chain (`cancelOnFirstWriteWriter` → `JSONFieldOrPlainStreamWriter` → `streamSessionWriter`) down to the minimax session's `sentenceBuffer`.
- Invoke it from `EnableFinalStreaming` so each streaming turn starts with a clean buffer.

**Timing:** reset happens at turn *start*, not end. The final answer's own buffered tail isn't flushed until `closeAndWait` runs after `runtime.Run` returns, so resetting on `DisableFinalStreaming` would have clipped the last sentence of legitimate answers. Resetting on `EnableFinalStreaming` only clears the *previous* turn's residual.

Adapters without `ResetBuffer` (volcengine, alicloud, fishaudio) degrade to the previous behavior.

## Tested

- `go build ./...`, `go vet` — clean
- `go test ./internal/agent/ ./internal/agent/tts/...` — all pass
- New tests:
  - `buffer_test.go` — `sentenceBuffer.Reset` drops sub-threshold residual and a later Write is uncontaminated
  - `stream_buffer_reset_test.go` — `ResetBuffer` propagates through the full writer chain to the terminal sink; safe no-op when unsupported

Note: the separate `INTERNAL_ERROR` from `SpeakToolContent`/streaming TTS contending for the AO device is a distinct concurrency issue and is intentionally out of scope here.

### DIFF
--- a/src/agent/internal/agent/json_field_stream.go
+++ b/src/agent/internal/agent/json_field_stream.go
@@ -405,6 +405,18 @@ func (w *JSONFieldOrPlainStreamWriter) ResetStreamState() {
 	w.emitted = false
 }
 
+// ResetBuffer forwards a buffer reset to the underlying target writer. The
+// JSON-field extraction layer itself holds no synthesizable buffer, but the
+// downstream TTS session may.
+func (w *JSONFieldOrPlainStreamWriter) ResetBuffer() {
+	if w == nil {
+		return
+	}
+	if resetter, ok := w.target.(ttsBufferResetter); ok {
+		resetter.ResetBuffer()
+	}
+}
+
 func (w *JSONFieldOrPlainStreamWriter) StreamEmitted() bool {
 	if w == nil {
 		return false
--- a/src/agent/internal/agent/progress_speaker.go
+++ b/src/agent/internal/agent/progress_speaker.go
@@ -190,6 +190,16 @@ func (w *cancelOnFirstWriteWriter) ResetStreamState() {
 	w.mu.Unlock()
 }
 
+// ResetBuffer forwards a buffer reset to the inner writer if it supports it.
+func (w *cancelOnFirstWriteWriter) ResetBuffer() {
+	if w == nil {
+		return
+	}
+	if resetter, ok := w.inner.(ttsBufferResetter); ok {
+		resetter.ResetBuffer()
+	}
+}
+
 func (w *cancelOnFirstWriteWriter) StreamEmitted() bool {
 	if w == nil {
 		return false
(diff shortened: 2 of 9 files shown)


icodeface:
## Independent review follow-up

An independent code review found one blocking gap, now fixed in db84450.

**Gap:** `finalStreamFanoutWriter` (used by the streaming chat path `handleChatStream`, server.go:1301) implemented `ResetStreamState` but not `ResetBuffer`. So on that path `resetStreamBuffer` type-asserted against the fanout writer, failed, and silently no-opped — the minimax `sentenceBuffer` was never cleared and the duplicate-speech bug persisted whenever streaming TTS was enabled there. The hardware voice path (`audio_dialog.go`) and the async/sync chat paths wrap `cancelOnFirstWriteWriter` directly, so they were already fixed by the original commit.

**Fix:** Added `ResetBuffer` to `finalStreamFanoutWriter`, mirroring its existing `ResetStreamState` fan-out, plus a test (`TestResetStreamBufferReachesTerminalThroughFanout`) asserting the reset reaches the TTS leg through the fanout.

### Review items verified clean (no change needed)
- **Reset timing**: confirmed reset happens at turn *start* (`EnableFinalStreaming`), before any write; the final answer's buffered tail is flushed only by `closeAndWait` → `Close` → `Flush` after `runtime.Run` returns, so legitimate content is never clipped.
- **Concurrency**: `sentenceBuffer.Reset` takes the same `buf.mu` as `Write`/`Flush`; `streamSessionWriter.ResetBuffer` reads the session under `w.mu` mirroring `Write`. No race.
- **Other adapters** (volcengine, alicloud, fishaudio) push text upstream immediately and hold no local sentence buffer, so the `ResetBuffer` no-op is correct for them. Minimax is the only client-side buffering adapter.

### Acknowledged, out of scope
- `finalTokenSeen` is not cleared in `DisableFinalStreaming` (pre-existing). In practice `EnableFinalStreaming` resets it at every turn start before `HasFinalStreamingToken` is read, so it does not affect this fix.

All tests pass: `go build ./...`, `go vet`, `go test ./internal/agent/ ./internal/agent/tts/...`.

PR #307, opened 2026-06-25
feat(agent): support recall archived sessions

ljunb:
## Summary

Closes the two long-term-memory gaps we found around session rotation:

1. **Archived sessions were unrecallable.** After ~30 minutes of inactivity, sessions rotated into `session_archive/` and no tool could read them back. The agent would simply tell the user it could not find earlier content.
2. **`MemoryCandidates` were discarded.** The LLM-generated structured summary already identified durable facts worth keeping (`memory_candidates`), but nothing promoted them to long-term memory before the chunks were archived.

This PR closes both, and tightens the tool description so the agent actually uses the new capability.

### 1. `recall_session_chunks` now searches archived sessions

- New `ArchivedSessionStore` scans `memory/session_archive/` and reuses `SessionMemoryStore` per archived directory for chunk lookup
- `ChunkRecallQuery` gains `include_archived` and `archived_time_range` (`last_24h` default, `last_7d`, `all`)
- `ChunkRecallResult` gains `session_id` so the agent can tell active vs archived hits apart
- Backward compatible: defaults stay active-only

### 2. Auto-promote memory candidates during rotation

- In `RotateSessionEventsDetailed`, before the directory rename, we read every active chunk's `Structured.MemoryCandidates` and persist each unique candidate as a `MemoryItem` (`type=fact`, `priority=70`)
- Tags, entities, and the chunk summary travel along as evidence
- Extraction is non-fatal — failures only log warnings

### 3. Stronger guidance in the tool description

- Wider trigger-word list (`yesterday`, `last time`, `previously`, `just now`, `a moment ago`, …) with explicit ''non-English equivalents count'' note
- New **RETRY RULE**: after an empty active-session search, the agent MUST retry with `include_archived=true` before claiming the content is missing
- Explicit reminder that sessions rotate after ~30 min, so even ''just now'' references can already be archived
- `archived_time_range` guidance: start narrow (`last_24h`), widen only if needed

## Testing

- `go test ./internal/agent -run 'TestRecallSessionChunks|TestRecallToolDescriptions|TestSessionRecallTelemetry|TestToolSetRegisters'` passes
- Extended `TestRecallToolDescriptionsGuideAgentUsage` to assert the new trigger words and RETRY RULE stay in the prompt
- Updated all existing callers of `NewRecallSessionChunksTool` to the new two-arg signature

## Files

- New: `session_archive_recall.go`, `session_memory_candidates.go`
- Modified: `memory.go`, `memory_tools.go`, `session_memory.go`, `tool_input.go`, `tools.go` (+ test updates)

## Summary by CodeRabbit

* **New Features**
  * Memory recall now searches both current and archived session history, including returning archived matches when available.
  * Recalled chunks can include the originating session ID when available.
* **Bug Fixes**
  * Session archiving now attempts a final “pre-archive” compression of remaining events and does not block rotation if compression fails.
  * Recalled chunks default to an `active` source when source info is missing; archived recall failures no longer prevent returning active results.
* **Tests**
  * Expanded coverage for archiving, empty-session behavior, and updated recall/telemetry expectations.


ljunb:
@coderabbitai review

ljunb:
@coderabbitai review

PR #323, opened 2026-06-29
fix: restore reliable iOS HID modifiers

ZhaoQing419:
## Summary
- keep the keyboard HID function in boot-keyboard mode (`protocol=1`, `subclass=1`) for broad host compatibility
- refresh the full USB composite gadget on stale host sessions instead of resetting ECM alone
- make the Agent reopen stale HID device descriptors after `/dev/hidg0` or `/dev/hidg1` is recreated, or after the USB watchdog records a composite refresh

## Root Cause
The iOS modifier failures were traced to stale HID state rather than the keyboard protocol itself.

A composite gadget reset can delete and recreate `/dev/hidg0` while the Agent process still holds an older open file descriptor. In that state, tool calls can return `ok` because the write to the old descriptor succeeds or does not fail clearly, while the current USB HID session on the phone does not reliably consume the report. The visible symptom is that normal typing may appear to work in some test paths, but modifier-based actions such as `Shift+1`, `Cmd+A`, and `Cmd+V` fail.

The current direction is therefore:
- keep `protocol=1/subclass=1` as the default keyboard profile
- use full composite re-enumeration to clear stale host-side sessions
- have the Agent verify and reopen HID descriptors before writing reports

## Changes
- `S49usbhid`: preserves boot keyboard `protocol=1/subclass=1` and re-enumerates the composite gadget after initial bind.
- `S60usb_ecm_watchdog`: records structured refresh state in `/run/aiden_usb_ecm_watchdog.state` and resets the full composite gadget on stale host/ECM conditions.
- Agent HID writer: before each HID report, checks whether the open descriptor still matches the current device node and whether the watchdog refreshed the gadget after the descriptor was opened. If stale, it closes and reopens the HID node.
- Tests: add focused checks for keyboard protocol, watchdog composite recovery, stale HID node recreation, and watchdog-triggered descriptor reopen.

## Verification
- `scripts/test_usb_gadget_watchdog.sh`
- `scripts/test_usb_hid_keyboard_protocol.sh`
- `cd src/agent && go test ./internal/agent -run 'TestHIDDevice|TestKeyboard|TestQuickAction|TestTapPointer|TestTouch|TestMouse'`
- `cd src/agent && go test ./internal/agent`
- `git diff --check`

## Hardware Notes
- Current observed board state after reboot: `hid.usb0 protocol=1`, `subclass=1`, `report_length=8`, UDC configured.
- Agent held a clean `/dev/hidg0` descriptor after reboot, and iOS paste worked in that state.
- Earlier `0/0` success is now understood as likely coupled with agent restart / clean descriptor refresh, not proof that report protocol is required by default.


HuangHan96 (changes requested):
需要保持 protocol=1/subclass=1 以保证最大的兼容性

HuangHan96 (approved):

PR #329, opened 2026-06-30
fix: streaming TTS drain blocking chat completion

web3miaomiao:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat streaming so speech output is tracked more reliably during both live and async responses.
  * Fixed cancellation and cleanup behavior so streamed audio can finish or stop without delaying the main chat response.
  * Added a fallback for cases where speech output does not emit during streaming, ensuring text still appears as expected.

* **Tests**
  * Expanded coverage for streaming chat, async completion, cancellation timing, and TTS fallback behavior.

### DIFF
--- a/src/agent/internal/agent/server.go
+++ b/src/agent/internal/agent/server.go
@@ -1019,13 +1019,7 @@ func (s *Server) handleChatAsync(
 
 		result, err := s.runtime.Run(runCtx, runReq)
 		progress.Cancel()
-		if newStream != nil {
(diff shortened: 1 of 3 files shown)


HuangHan96:
看了下，真正的问题应该是 TTS streaming 输出时 drain 的时间计算错误，导致最后等待了完整音频的长度

HuangHan96 (changes requested):
需要修复 drain 时长计算错误的问题

PR #344, opened 2026-07-02
fix(agent): isolate recoverable loop errors

ljunb:
## Summary
- Treat tool-local context errors as structured tool observations while preserving hard aborts for parent run cancellation/deadlines.
- Apply the same parent-cancellation check to environment bridge tool results.
- Make memory/session load, begin, context-view, and commit failures best-effort so they do not hide model output when the run context is still alive.

## Test Plan
- go test -count=1 ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved run resilience: failures loading persisted memory, session start/commit, or conversation history no longer block completion when the request is still active.
  * Session context loading now respects cancellation: when canceled, it returns the cancellation; otherwise it continues with already-available values.
  * Tool execution now consistently treats cancellation/deadlines as the controlling outcome, ensuring correct structured cancellation behavior.
* **Tests**
  * Added resilience tests for continued successful runs when persistence/session operations fail.
  * Expanded tool execution tests for local and parent-context cancellation, including cancellation during environment bridging.


ljunb:
不好意思，忘了 update  branch了，我处理下先

HuangHan96 (approved):

PR #369, opened 2026-07-07
fix(agent): parse frame health string fields

web3luhao:
## Summary
- parse frame service health numeric fields through the same flexible string-or-number path used by frame metadata
- keep android-adb status from misclassifying a healthy frame service when the health response serializes counters as strings
- add a regression test covering string-encoded health fields

## Testing
- go test ./internal/agent -run 'TestFrameMetadataUnmarshalSupportsStringNumbers|TestFrameMetadataUnmarshalAllowsOmittedSourceAndCropFields|TestFrameHealthResponseSupportsStringNumbers|TestAndroidADBStatusMatchesCurrentPhoneUSBIP|TestAndroidADBStatusRequiresPairWhenFrameUnavailableAndDeviceMismatched'

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved health parsing to accept numeric fields returned as either numbers or strings.
  * Health reporting now consistently preserves state, last error, and latency averages even when formats vary across fields.

* **Tests**
  * Added coverage to ensure string-formatted numeric health values decode correctly and report the expected results.


HuangHan96 (changes requested):
只需要把 frame service 和 frame client 统一下 response 的字段类型就行，没必要专门做额外的两种格式的兼容。

PR #424, opened 2026-07-20
fix(agent): stop touch_gesture retry/oscillation loops

ljunb:
## Summary

Fixes two independent root causes behind the `touch_gesture` loop QA reported (model repeatedly tapping without progress).

### 1. Stop advertising `screenshot` coord_space
`screenshot` is effectively the renamed `pixel` space removed in #387 — it maps to pixels inside the cropped active-area screenshot. #398 reintroduced it into the shared `coordSpaceSchema` while adding the wheel picker, so it leaked back onto every pointer tool. Small models then pass cropped-screenshot pixels whose x overflows the active-area width (e.g. x=900 against a 496px-wide mirror), failing the bounds check and driving retry loops.

- Removed `screenshot` from the coord_space enum and the mouse/touch tool descriptions; steer toward normalized 0-1000.
- Internal resolver still accepts `screenshot` for backward compatibility (interpreted, not hard-failed).
- `wheel_nudge` never exposed coord_space — unaffected.

### 2. Detect oscillating-screen loops in the termination policy
The loop guard treated any screenshot-hash change as progress and reset its stall counters. When a repeated tap toggled the keyboard on/off, the screen alternated A-B-A-B, every step looked like progress, and **31 identical taps over 8 minutes never tripped any tier of the guard** (confirmed in the QA logs — no notice/restrict/terminate ever fired).

- Track a bounded window of recent screen hashes; revisiting a recently seen screen no longer counts as progress.
- A repeated action that only oscillates between known screens accrues stall like a static screen does.
- `recordProgress` keeps the window (so oscillation stays detectable); `ResetForSteer` clears it.
- Result: A-B-A-B oscillation hits soft-notice on the 3rd repeat and terminates on the 4th, instead of looping unbounded.

## Testing
- `go test ./internal/agent/` — pass (added oscillation-detection and genuinely-new-screen tests; existing 5 termination tests unchanged).
- `go build ./...`, `go vet ./internal/agent/` — pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved termination/loop detection to stop when the interface oscillates between recently seen screens without real progress, while continuing when screens are genuinely new.
  * Added explicit oscillation handling for loop-restricted UI action scenarios.
  * Preserved backward compatibility for screenshot-based coordinate inputs.

* **Improvements**
  * Updated coordinate-space guidance and schemas to favor normalized `0–1000` coordinates with `(500,500)` as center (removing `"screenshot"` from model-exposed options).

* **Tests**
  * Added coverage for oscillating vs. genuinely new screen observations and updated HID tool schema expectations.


src/agent/internal/agent/termination_policy.go
```diff
@@ -256,6 +264,13 @@ func (p *TerminationPolicy) AfterToolCall(toolName, input, observation string, i
 	sameSignature := signature != "" && signature == previousSignature
 	madeProgress := false
 
+	// Revisiting a screen seen in the recent window is oscillation, not
+	// progress: a repeated action toggling the UI between a few states keeps
+	// changing the hash but goes nowhere. Computed up front so both the
+	// result-hash and screen-hash progress checks below respect it.
+	screenHash := extractScreenshotDataHash(observation)
+	revisitedScreen := screenHash != "" && p.seenRecentScreen(screenHash)
+
```
HuangHan96: 用 screen-hash 是没法确认屏幕内容是不是相同的，这个没什么用，可以先去掉，后面单独找方案解决 A/B/A/B 问题

HuangHan96 (changes requested):

src/agent/internal/agent/termination_policy.go
ljunb: 已按建议在本 PR 去掉（revert commit d8ec3ad4）。screen-hash 确实无法可靠判断屏幕内容是否相同，A/B/A/B 循环后续单独用更强的信号（如 image_diff 结构化 screen_changed / 动作签名硬上限）另开方案解决。本 PR 只保留 coord_space 修复。

HuangHan96 (approved):

PR #441, opened 2026-07-24
fix(hid): address PR 440 review gaps

ZhaoQing419:
## Summary

- make iOS HID isolation ownership context-cancelable so preempted runs do not remain blocked behind detached HTTP tools
- batch direct HTTP run_script modifier operations across the whole script
- prefer runtime Phone Bridge/config platform information before the iOS controller fallback in search_launch_app
- add focused regression tests for all three cases

## Verification

- go test ./...
- go test ./internal/agent -count=1
- focused go test -race coverage
- go vet ./...
- sh scripts/test_dynamic_keyboard.sh
- sh scripts/test_usb_gadget_watchdog.sh
- git diff --check

Stacked on #440.


ZhaoQing419:
Fix commit was pushed directly to #440 as requested; closing this stacked PR.

PR #449, opened 2026-07-27
fix(agent): resolve image diff from screenshot attachments

ZhaoQing419:
## Summary

- expose persisted screenshot attachment filenames to the model as `screenshot_attachment_id`
- allow Agent `image_diff` calls to resolve `before` and `after` from screenshot attachments in the active context
- retain Base64 JPEG compatibility for direct HTTP `image_diff` calls
- reject arbitrary paths, non-screenshot attachments, and attachment IDs unavailable in the active context
- keep the existing image comparison algorithm and change threshold unchanged
- update Agent documentation and device-operation guidance to use screenshot attachment IDs

## Verification

- `go test ./internal/agent/...`
- `go vet ./internal/agent/...`
- `git diff --check`
- GitHub Actions CI passed


HuangHan96 (changes requested):
不需要另一套 screenshot id 机制，context manager 中 attachment 文件名已经是给每个 screenshot 独特的 id 了；可能需要在转换为标准 llm messages 保留文件名信息，让 agent 在上下文中能够感知这个文件名信息

HuangHan96 (approved):

PR #496, opened 2026-08-10
fix(agent): share TTS provider manager across voice paths

icodeface:
## Summary

- make Runtime the single stable owner of the TTS provider manager shared by HTTP Server and AudioDialog
- bind provider capabilities, playback sink format, and stream session to the same provider generation during hot switches
- serialize provider switches with manager shutdown and reject switches after close
- inject provider factories for deterministic tests without mutating the global registry
- prove both configured hot switches and first-time enablement through POST -> AudioDialog.Speak -> provider session/text

## Tests

- go test ./... -count=1
- go test -race ./internal/agent/tts ./internal/agent -run targeted-TTS-regressions -count=5
- go vet ./...
- git diff --check

## Summary by CodeRabbit

- **New Features**
  - Audio conversations and server speech now share the active text-to-speech provider.
  - Switching TTS providers updates audio behavior consistently without restarting the application.
  - Audio streaming adapts to the selected provider’s capabilities.

- **Bug Fixes**
  - Improved handling when no TTS provider is configured.
  - Prevented active audio sessions from being disrupted during provider changes.
  - Improved interruption and cleanup of audio streams during shutdown.
  - Servers now reject new requests during shutdown and allow active requests to finish.
  - Improved reliability when changing or closing providers during active audio sessions.


icodeface:
Addressed the review feedback in 86208f36. AudioDialog now closes only dialog-owned resources, and the regression test verifies that closing the dialog does not close the Runtime-owned TTS provider; Runtime.Close still closes it exactly once. The switched-provider speech-path assertion was already covered by 0bdacdb1 and remains passing.

PR #611, opened 2026-08-31
fix(benchmark): resolve mac desktop health check path

codesister:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS screen-size detection by locating the system profiling utility across supported installation paths.
  * Increased compatibility on systems where the utility is not installed in its previously expected location.
  * Preserved accurate screen-resolution reporting when using an alternate utility location.

* **Tests**
  * Added coverage verifying utility discovery, fallback behavior, command execution, and screen-resolution parsing on macOS.

### DIFF
--- a/benchmark/desktop/bridge/device.py
+++ b/benchmark/desktop/bridge/device.py
@@ -91,7 +91,11 @@ class DesktopDevice:
             size = self._pyautogui.size()
             return int(size[0]), int(size[1])
         if self.system == "darwin":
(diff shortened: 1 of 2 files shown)


HuangHan96 (approved):
Reviewed the macOS executable discovery and fallback behavior. The change correctly prefers PATH discovery, falls back to the standard /usr/sbin location, and includes focused regression coverage for that fallback. No blocking issues found.\n\nValidation: uv run pytest tests/test_desktop_bridge.py -q -p no:capture (13 passed).

PR #616, opened 2026-09-01
feat(agent): add provider-aware reasoning controls

ljunb:
## Summary

- add provider-neutral reasoning capability metadata and models.dev enrichment
- expose model-specific reasoning controls through `/api/models` and Config Web
- map reasoning effort and exact budgets to Anthropic adaptive/manual thinking
- preserve signed Anthropic thinking blocks across tool-call turns
- raise the default max response limit to 8192 tokens

## Configuration

```toml
[model]
reasoning_effort = "high"
reasoning_budget_tokens = 4096
max_response_tokens = 8192
```

`reasoning_budget_tokens` is currently translated only to Anthropic's native `thinking.budget_tokens`. Aiden-owned APIs and configuration consistently use `reasoning`.

## Validation

- `GOCACHE=/tmp/aiden-go-cache-thinking go test ./... -run '^$'`
- focused model metadata, `/api/models`, config, Anthropic reasoning, and config wire tests
- Config Web JavaScript syntax checks
- `git diff --check`

Benchmark suites were not added because this changes configuration, capability discovery, and provider protocol behavior rather than benchmark task behavior.

## Summary by CodeRabbit

* **New Features**
  * Added configurable realtime voice providers, routing, authentication, and provider-specific settings.
  * Model details now include reasoning capabilities, API information, limits, and custom-model metadata.
  * Added reasoning token budgets and adaptive model-specific reasoning controls.
  * Reasoning content is now preserved and displayed separately in chats, streaming responses, voice runs, and LLM logs.
  * Changed the default input mode from voice transcription to text.
* **Bug Fixes**
  * Truncated responses now return clear errors without retries.
  * Invalid saved configurations can be repaired through the configuration interface.
* **Documentation**
  * Expanded guidance on reasoning, providers, authentication, and output limits.


src/agent/internal/agent/thinking_text.go
```diff
@@ -0,0 +1,254 @@
...
+			}
+			return nil
+		case final:
+			if s.pending.Len() > 0 {
+				if err := s.emit(s.pending.Bytes()); err != nil {
+					return err
+				}
+				s.pending.Reset()
+			}
+		default:
```
icodeface, line 210: **Visible content is never streamed incrementally once `reasoning_effort` is set, which stalls TTS for the whole response.**

`forwardUntagged` is wired as `m.reasoningEffort == "" && stream != nil` (openai_compatible_model.go), so any configured effort makes it `false`. For a provider that returns reasoning neither in think tags nor in a `reasoning_content` field (reasoning hidden server-side, which is the common case for OpenAI-style `reasoning_effort`), no chunk ever matches a case here:

- `s.inThinking` is false (no opening tag)
- `openAt`/`closeAt` are both `-1` (no tags at all)
- `s.foundTag` is false
- `s.forwardUntagged` is false
- `final` is false (see the separate note on `case final:`)

So every chunk falls through to this `default: return nil` and accumulates in `pending`. Nothing reaches `downstream` until `Finish()` flushes the entire response in one shot.

Reproduced by feeding `"Hello "`, `"there, "`, `"how "`, `"are "`, `"you?"` into a stream built with `forwardUntagged=false`:

```
visible chunks BEFORE Finish: 0 []
visible chunks AFTER  Finish: 1 ["Hello there, how are you?"]
```

Since the visible callback is what drives `speech.StreamWriter`, TTS gets nothing until the model finishes generating. Voice is the primary interaction mode for this device, so enabling reasoning effectively converts streaming speech into wait-for-the-whole-answer speech. Text mode only looks fine because the full flush arrives at the end.

The ambiguity this branch guards against is an *orphan closing tag*, which only matters until the first non-tag-prefix byte. Suggest forwarding untagged text whenever no opening tag has been seen and the buffer does not end in a potential tag prefix (the logic the `case s.forwardUntagged` branch above already implements), independently of whether an effort was configured.

src/agent/internal/agent/thinking_text.go
```diff
@@ -0,0 +1,254 @@
...
+
+// FlushVisible switches a stream back to ordinary content when the provider
+// has already supplied reasoning through its dedicated field. This avoids
+// delaying correctly separated visible deltas just because reasoning was
+// enabled for the request.
+func (s *taggedThinkingStream) FlushVisible() error {
+	if s == nil || s.downstream == nil || s.pending.Len() == 0 {
+		return nil
+	}
+	err := s.emit(s.pending.Bytes())
```
icodeface, line 136: **`FlushVisible` emits buffered reasoning bytes to the visible/TTS callback and clears `inThinking` in the middle of an open think block.**

While `inThinking` is true, `pending` holds a *partial closing-tag prefix* deliberately withheld by `process` — that content is reasoning, not visible text. `FlushVisible` sends it straight to `s.downstream` and then unconditionally sets `inThinking = false`.

Reproduced with an open think block whose closing tag is split across chunks, then a `reasoning_content` delta arriving (which is exactly what triggers `FlushVisible` in `decodeStreamingResponse`):

```
write: "<think>secret plan</thi"
  -> inThinking=true reasoning=["secret plan"] visible=[]
FlushVisible()
  -> visible=["</thi"]        // tag garbage forwarded to TTS
```

And the state reset, with `"<think>step one</"` buffered:

```
before flush: inThinking=true pendingLen=2
after  flush: inThinking=false visible=["</"]
```

Two consequences: a `</thi` fragment is spoken/rendered as visible content, and the remainder of a still-open think block is no longer classified by the `inThinking` path. In my trace the leftover reasoning happened to still be captured, but only because the orphan-close fallback (`case closeAt >= 0`) caught it — that is luck, not intent, and it breaks as soon as the block contains an opening tag for a nested/subsequent segment.

Suggest making `FlushVisible` a no-op (or routing `pending` to `emitReasoning`) when `s.inThinking` is true, and leaving `inThinking` untouched. The stated purpose — not delaying correctly separated visible deltas — only applies when the stream is not inside a think block.

src/agent/internal/agent/thinking_text.go
```diff
@@ -0,0 +1,254 @@
...
+			// delivered as soon as it is no longer a tag prefix.
+			safe := len(text) - maxThinkingTagPrefixLen(text, thinkingOpenTags)
+			if safe > 0 {
+				if err := s.emit(s.pending.Bytes()[:safe]); err != nil {
+					return err
+				}
+				s.pending.Next(safe)
+			}
+			return nil
+		case final:
```
icodeface: **`case final:` is dead code.** `process` has exactly one caller, `Write` at line 93, which passes `false`; `Finish()` does its own flushing and never calls `process`. So the `final` parameter is always `false` and this branch is unreachable.

It is not merely redundant: it looks like the intended terminal-flush path, which makes the `default: return nil` fall-through above easy to misread as safe. Either drop the parameter and this branch, or have `Finish()` call `process(true)` so the flush logic lives in one place.

icodeface (commented):
Code review of the reasoning/thinking control changes, focused on provider parameter mapping, streaming order of reasoning vs. visible text, config defaults/back-compat, and voice (TTS) latency.

Three issues found, all in the new `taggedThinkingStream`. Each was reproduced by running the new file in isolation against synthetic provider streams, so the behaviour below is observed rather than inferred. The provider parameter mapping (`reasoning_effort` passthrough, OpenRouter `reasoning.exclude`, Anthropic adaptive vs. `budget_tokens`), the `budget >= maxTokens` clamping, the thinking-block replay/hoisting, and the config defaults all checked out.

icodeface (dismissed):

aiden-lightning:

- PR: #616 — feat(agent): add provider-aware reasoning controls
- Head SHA: `cf70767fe27d86f90c52f437ac325e9bc470f2f7`
- GitHub updatedAt: `2026-09-04T03:51:34Z`
- Reviewed at: `2026-09-04T12:55:23+08:00`

## PR #616 Review: Provider-Aware Reasoning Controls

### Summary

Reviewed 7 highest-risk files from the 63-file changeset, focusing on the core Anthropic reasoning implementation (`anthropic_model.go` +305/-10), metadata fetching (`model_metadata.go` +218/-3), and configuration validation (`config.go` +62/-14). The PR adds native Claude extended thinking support with budget controls and extends provider metadata fetching to models.dev for reasoning capability discovery.

**Verified defects:** 1 confirmed  
**Concerns requiring clarification:** 1

---

### Coverage

**Files reviewed (7 of 63):**
- `src/agent/internal/agent/anthropic_model.go` (patch + WebFetch verification)
- `src/agent/internal/agent/model_metadata.go` (patch + base file + WebFetch verification)
- `src/agent/internal/agent/model_specs.go` (base file read)
- `src/agent/internal/agent/config.go` (patch via WebFetch)
- `src/agent/internal/agent/model/model_spec.go` (base file read)
- Test files: `anthropic_thinking_budget_test.go`, `config_test.go` (patch samples)

**Not reviewed:** 56 changed files including runtime.go, models.go, all OpenAI reasoning implementations, UI components, and 27 files whose patches were omitted from context.

---

### Verified Issues

#### 1. models.dev catalog errors cached permanently (medium severity)

**File:** `src/agent/internal/agent/model_metadata.go`  
**Evidence:** WebFetch confirmed the cache implementation stores errors without eviction

The new `modelsDevCatalogCache` uses a `sync.Map` with entries containing `{ready chan struct{}, catalog, err}`. When a catalog fetch fails (network timeout, DNS failure, or API error), the code stores the error in the cache entry and closes the ready channel:

```go
entry.err = err
close(entry.ready)
```

The entry is never deleted from the sync.Map. Subsequent lookups for that URL wait on the closed channel and read the stored error repeatedly.

**Impact:** A transient network failure at daemon startup (e.g., brief DNS timeout, models.dev rate limit) poisons metadata for anthropic/openai/volcengine/kimi providers for the entire process lifetime. Every model instantiation requiring reasoning metadata will fail with the cached error until the daemon restarts.

**Recommendation:** Either (a) delete the cache entry on error so the next attempt retries, or (b) add TTL/expiry so transient errors self-heal after N minutes without requiring a daemon restart.

---

### Concerns Requiring Clarification

#### Cache version not bumped despite schema change

**File:** `src/agent/internal/agent/model_metadata.go`  
**Lines:** Cache version constant remains `1`

The PR adds a `Reasoning *ReasoningCapability` field to `model.ModelSpec` and updates `hasProviderModelSpecMetadata()` to check `spec.Reasoning != nil` as a validity condition:

```go
// Post-PR
return spec.ContextWindow > 0 || spec.MaxOutput > 0 || spec.Reasoning != nil
```

However, `providerModelMetadataCacheVersion` was not incremented from `1`. Cache files written before this PR will have specs without the `Reasoning` field (nil). When the upgraded daemon reads these entries:

1. For openrouter/ollama: No issue, since `needsReasoning` explicitly excludes them
2. For the newly-added providers (anthropic/openai/volcengine/kimi): The cached spec has `Reasoning == nil`, but the logic now treats that as "needs fetch"

**Possible outcomes:**
- The cache entry might be considered valid (ContextWindow > 0), but `needsProviderModelMetadataForSpec` will still request reasoning metadata, causing a redundant fetch
- Or the entry is rejected and a fresh fetch occurs anyway

**Question:** Is this intentional (cached entries are provider-specific and the new providers have no old cache), or should the version be bumped to `2` to force a clean slate?

---

### Verdict

**HOLD pending fix for Issue #1.** The permanent error caching is a regression that affects daemon reliability in environments with transient network issues. The validation logic, budget enforcement, and nil-spec fallback behavior are all correctly implemented based on the available evidence.

HuangHan96 (changes requested):
应该不用改动 pico-sdk 吧

ljunb:
> 应该不用改动 pico-sdk 吧

我确认下

PR #627, opened 2026-09-02
refactor(agent): remove obsolete session-store machinery

HuangHan96:
## Summary

Removes the session-store layer, which logged conversation events before compaction. This layer is now obsolete since conversation chunks are written directly by the context manager at compaction time.

## What was removed

**Core session store components:**
- `SessionMemoryStore` and `SessionEvent` types
- `AppendEvent`, `Compress`, `RotateSessionEvents` methods
- `ArchivedSessionStore` for reading old archives
- Session-specific compaction config (summary templates, chunk size limits)
- Session boundary detection logic (`sessionBoundaryTelemetry`, `recentEpisodeContext`)

**Files deleted:**
- `storage_write_gate_test.go` and 60+ obsolete tests

## What changed

- `RegisterMemoryTools` signature: `(configDir, memoryDir string, store)` 
- `RecallSessionChunksTool` now wraps `MultiSessionChunkStore` (reads chunks directly from context manager sessions)
- Removed 12 obsolete config keys from schema validation
- Fixed `countRecalledChunks` to only count pending-source results for telemetry

## What remains

- `MultiSessionChunkStore` for reading conversation chunks across sessions
- Long-term memory tools (`save_memory`, `recall_memory`, `forget_memory`)
- Device memory and episode inspection tools
- Context manager (the actual conversation history store)

## Verification

✓ All builds succeed
✓ All non-audio tests pass (audio failures are environmental socket path issues)
✓ Core memory tests verified: recall, save, forget, episodes all working
✓ Net removal: ~9,400 lines deleted, ~800 lines added

[Open in Delta](https://delta.dev/join/thread_01m1gbwh2y73hymxepp30fxda5)

## Summary by CodeRabbit

- **New Features**
  - Session memory recall now searches across multiple sessions.
  - Memory chunks support searchable tags and entities derived from conversation content.
  - Compacted conversation chunks can be persisted for later recall.
  - Session and HTTP log handling now uses the current context session.

- **Behavior Changes**
  - Legacy session-compaction and session-boundary settings are ignored.
  - Heuristic session-boundary detection and automatic rotation are no longer performed.

- **Reliability**
  - Safer session-based log filenames and improved handling of truncated session data.
  - Concurrent chunk writes are coordinated to preserve indexed memory.


src/agent/internal/agent/compactor/compactor.go
```diff
@@ -134,6 +139,14 @@ func (c *Compactor) Compact(ctx context.Context, session *contextmanager.Context
 	if summary == "" {
 		return nil, false, fmt.Errorf("failed to generate summary")
 	}
+
+	// Write chunk for the compacted messages before creating revision
+	if chunkWriter != nil {
```
icodeface, line 144: **Chunk written before compaction completes; partial failure creates duplicate chunks**

WriteChunk succeeds (chunk file and index entry created), then session.CreateRevision fails (e.g., disk full). The compacted messages remain in the transcript. Next compaction attempt compacts the same messages again, creating a second chunk. recall_session_chunks returns both chunks, showing duplicate history.

Consider moving WriteChunk after successful CreateRevision, or making the operation atomic/idempotent.

src/agent/internal/agent/session_chunk_writer.go
```diff
@@ -0,0 +1,194 @@
...
+	if w == nil || w.sessionFolder == "" || sessionID == "" {
+		return nil
+	}
+	if len(msgs) == 0 {
+		return nil
+	}
+
+	// One writer at a time per process: the index below is a read-modify-write,
+	// so concurrent compactions would otherwise drop each other's entries.
+	w.mu.Lock()
```
icodeface: **Process-level mutex doesn't protect cross-process concurrent chunk writes**

Two daemon processes compact simultaneously (e.g., after a crash and restart, or two instances running). Both read chunks/index.yaml, both append an entry, both write back. Last write wins; one chunk entry is lost. Chunk file exists but is not in index, making it unreachable.

Consider using a file lock (like the old FileLock) to coordinate across processes.

src/agent/internal/agent/runtime.go
```diff
@@ -1343,6 +1308,20 @@ func (r *Runtime) getSystemPrompt() string {
 	return profile.SystemPrompt
 }
 
+// sessionChunkWriter builds the writer that persists a compacted conversation
+// span as a searchable chunk. Compaction is the only producer of chunks, so both
+// the threshold and overflow-recovery paths share this one construction.
```
icodeface, line 1313: **sessionChunkWriter creates new instance per call; mutex doesn't protect concurrent compactions**

If run gate allows concurrent runs or is disabled, two runs compact simultaneously. Each gets a separate SessionChunkWriter instance with a separate mutex. Both read chunks/index.yaml (via WriteChunk), both append an entry, both write back. Last write wins; one chunk entry is lost.

Consider caching a single SessionChunkWriter on the Runtime, or using a file lock.

src/agent/internal/agent/session_chunk_writer.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil
+	}
+	return out
+}
+
+// WriteChunk writes a chunk of messages with their summary to the session's chunks directory.
+func (w *SessionChunkWriter) WriteChunk(ctx context.Context, sessionID string, msgs []messages.Message, summary string) error {
+	// A nil writer is wrapped in a non-nil ChunkWriter interface by the caller,
+	// so guard the receiver here rather than relying on an interface nil check.
+	if w == nil || w.sessionFolder == "" || sessionID == "" {
```
icodeface, line 77: **WriteChunk doesn't check context cancellation; can leave partial chunk files**

Context canceled after file.Create but before file.Sync. JSONL file is partially written and incomplete. Next recall attempt reads corrupt JSON from the partial file and fails, or succeeds with truncated messages.

Add `if err := ctx.Err(); err != nil { return err }` at the start and potentially between long operations.

src/agent/internal/agent/multi_session_chunk_store.go
```diff
@@ -0,0 +1,124 @@
...
+		if _, err := os.Stat(indexPath); os.IsNotExist(err) {
+			continue
+		}
+
+		// Load and search this session's chunks
+		index, err := loadChunkIndexFromPath(indexPath)
+		if err != nil {
+			continue // Skip sessions with corrupt indices
+		}
+
```
icodeface, line 64: **Corrupt chunk indices are silently skipped without logging**

A chunk index.yaml is corrupted (bad YAML syntax, truncated file). RecallChunks skips that session silently. User's historical chunks from that session are unreachable, but no error or log indicates the problem.

Consider logging a warning when a chunk index fails to load (though the store has no logger field currently).

src/agent/internal/agent/multi_session_chunk_store.go
```diff
@@ -0,0 +1,124 @@
...
+				}
+				if chunk.Structured != nil {
+					result.Structured = chunk.Structured
+				}
+				allResults = append(allResults, result)
+			}
+		}
+	}
+
+	// Apply limit
```
icodeface, line 81: **Recall results truncated to limit without sorting; returns arbitrary chunks**

100 chunks match the query, limit=10. The first 10 chunks in filesystem directory order are returned (arbitrary, depends on directory implementation). Most recent or most relevant chunks might be skipped.

Consider sorting results by timestamp or relevance before truncating, or document that the limit returns arbitrary matches.

src/agent/internal/agent/openai_compatible_model.go
```diff
@@ -239,10 +239,8 @@ func (l *llmRawHTTPLogger) Log(ctx context.Context, entry RawHTTPLogEntry) error
 	}
 
 	fileName := "llm-http-" + fileTime.Format("200601021504") + ".log"
-	if sid := strings.TrimSpace(sessionID); sid != "" {
-		if err := validateSessionIDPathComponent(sid); err == nil {
-			fileName = "llm-http-" + sid + ".log"
-		}
+	if sid := strings.TrimSpace(sessionID); sid != "" && safeLogPathComponent(sid) {
```
icodeface, line 242: **Invalid session ID for log filename falls back silently without warning**

Runtime passes a session ID with path separators or control chars (bug in CurrentSessionID). safeLogPathComponent returns false, log filename falls back to timestamp. Developer trying to correlate logs with sessions can't find the session's log file and has no indication the session ID was rejected.

Consider logging a debug/warning message when falling back.

src/agent/internal/agent/session_chunk_writer.go
icodeface: Fixed by c927e038. The per-instance `sync.Mutex` is gone and the index read-modify-write now takes an `flock` on the `chunks` directory via `NewFileLock`. Since flock is held per open file description, this covers a second daemon process sharing the config dir as well as two goroutines in one process. Resolving.

src/agent/internal/agent/runtime.go
icodeface: Fixed by c927e038. Moving the lock off `SessionChunkWriter` onto the chunk index addresses this directly — a fresh writer per compaction no longer means a fresh lock, so concurrent compactions in one process now serialize. `TestSessionChunkWriterSerializesConcurrentWritersOnSameIndex` covers it by racing 8 independent writers against one session index. Resolving.

PR #628, opened 2026-09-02
fix(agent): retry truncated episode memory extraction

ljunb:
## Summary

- detect output-token truncation from provider stop reasons before parsing partial JSON
- reduce normal Episode Memory batches to three and retry truncated batches one Episode at a time with increasing output budgets
- apply the same adaptive truncation handling to omission review and retention audit calls
- keep potentially sensitive model response bodies out of persistent agent logs while retaining size diagnostics
- document the retry/recovery behavior and add regression coverage through the real OpenAI-compatible decoder

## Testing

- `go test ./internal/agent -run '^TestEpisodeMemory|^TestMemoryMerge' -count=1`
- `go test -race ./internal/agent -run '^TestEpisodeMemory(OmissionReviewTruncationUsesLargerRetryBudget|RetentionAuditTruncationUsesLargerRetryBudget|FailureLogDoesNotPersistResponseBody|TruncatedBatch)' -count=1`
- `go vet ./internal/agent`

Full `go test ./internal/agent` currently fails at `TestOpenAICompatibleModelDoesNotBufferRawHTTPResponseWithoutContextMarker` with `decode response: unexpected EOF`; the same isolated failure reproduces on `origin/main`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when AI responses are truncated, including automatic retries with increased limits.
  - Added recovery for retention and omission-review failures.
  - Improved handling of empty, invalid, and partially completed responses.
  - Reduced processing batch sizes to improve consistency and latency isolation.
  - Prevented unstarted retry groups from being marked as processing prematurely.

- **Privacy**
  - Failure logs record response sizes without storing response contents.
  - Provider-generated error details are sanitized before being logged or saved.

- **Reliability**
  - Preserved stale-revision recovery while allowing additional valid retry opportunities.


ljunb:
Addressed the unresolved review comment in commit `26a7d210`: provider-generated Episode Memory errors are now sanitized before logging or persisting `LastError`, while retention/omission/truncation sentinels remain available to `errors.Is` and retry classification. Added regression coverage for sensitive provider text.

src/agent/internal/agent/episode_memory.go
```diff
@@ -350,16 +386,18 @@ func (p *episodeMemoryProcessor) collectEpisodeMemoryWork(state *episodeMemorySt
 func (p *episodeMemoryProcessor) extractEpisodeMemoryWork(ctx context.Context, state *episodeMemoryStateFile, works []episodeMemoryWork, result *MemoryBatchResult) (bool, error) {
 	indexes := make([]int, 0, len(works))
 	episodes := make([]TaskEpisode, 0, len(works))
+	attempt := 0
 	for index := range works {
 		if works[index].needsModel {
 			indexes = append(indexes, index)
 			episodes = append(episodes, works[index].episode)
+			attempt = max(attempt, works[index].originalStatus.AttemptCount, works[index].status.AttemptCount)
```
icodeface: **Potential inefficiency**: Batch token budget sized for max attempt count, inflating budget for first-attempt episodes batched with retries.

**Scenario**: Episodes ep1(attempt=0) and ep2(attempt=2) both have omission-review failures (RetryBatchLimit=0, so they can batch). extractEpisodeMemoryWork computes `attempt=max(0,2)=2`. proposeEpisodeBatch requests episodeMemoryBatchTokenBudget(2, 2, 8000, 2)=higher budget than needed. ep1 wastes tokens on first attempt.

Consider sizing budget per-episode or documenting this shared-budget behavior.

icodeface: When episodes with different attempt counts are batched together (which can happen when they all have `RetryBatchLimit=0`, meaning non-truncation retries like omission-review failures), all episodes share the token budget sized for the maximum attempt count in the batch.

**Scenario:** Episodes `ep1` (attempt=0) and `ep2` (attempt=2) both failed omission review (not truncation, so `RetryBatchLimit=0`). They batch together, and line 394 computes `attempt=max(0,2)=2`. The batch then requests `episodeMemoryBatchTokenBudget(2, 2, ...)` which gives a larger budget than needed. Episode `ep1` on its first attempt gets an inflated token budget, wasting tokens.

While the `RetryBatchLimit` mechanism prevents batching first-timers with truncation retries, multiple non-truncation retries with different attempt counts can still batch together and unnecessarily inflate the budget for lower-attempt episodes.

src/agent/internal/agent/episode_memory.go
```diff
@@ -278,6 +300,20 @@ func (p *episodeMemoryProcessor) processBatchLocked(ctx context.Context, limit i
 }
 
 func (p *episodeMemoryProcessor) collectEpisodeMemoryWork(state *episodeMemoryStateFile, episodes []TaskEpisode, limit int, shouldStop func() bool) ([]episodeMemoryWork, MemoryBatchResult, error) {
+	// A truncated batch can leave every episode in the batch with the same
+	// failure. Persisting a one-episode retry limit on those statuses prevents
+	// the next worker pass from replaying the oversized batch; normal work keeps
+	// the configured batch limit.
+	for _, episode := range episodes {
+		status := state.Episodes[episodeMemoryStateKey(episode.ID, episodeMemoryExtractorVersion)]
+		if status.Status != episodeMemoryStatusRetry || status.RetryBatchLimit <= 0 {
```
icodeface: The code checks `status.RetryBatchLimit <= 0` to skip episodes without a retry batch limit, but doesn't validate that the value is actually meaningful (0 or positive).

**Scenario:** A bug elsewhere sets `episodeMemoryEpisodeStatus.RetryBatchLimit = -1`. The check evaluates to true (`-1 <= 0`) and continues without error or warning. The corrupt negative value is neither applied nor flagged, silently violating the batch limiting invariant.

Consider adding validation or using `== 0` if only positive limits are meaningful.

src/agent/internal/agent/episode_memory.go
```diff
@@ -278,6 +300,20 @@ func (p *episodeMemoryProcessor) processBatchLocked(ctx context.Context, limit i
 }
 
 func (p *episodeMemoryProcessor) collectEpisodeMemoryWork(state *episodeMemoryStateFile, episodes []TaskEpisode, limit int, shouldStop func() bool) ([]episodeMemoryWork, MemoryBatchResult, error) {
+	// A truncated batch can leave every episode in the batch with the same
+	// failure. Persisting a one-episode retry limit on those statuses prevents
+	// the next worker pass from replaying the oversized batch; normal work keeps
+	// the configured batch limit.
```
icodeface, line 307: **Missing validation**: Negative RetryBatchLimit silently bypasses limit adjustment instead of being rejected.

**Scenario**: A bug elsewhere sets episodeMemoryEpisodeStatus.RetryBatchLimit=-1. This check `status.RetryBatchLimit <= 0` continues without error or warning. The corrupt value is neither applied nor flagged, silently violating the batch limiting invariant.

Consider adding validation: `if status.RetryBatchLimit < 0 { return error }` or documenting that RetryBatchLimit must be >= 0.

ljunb:
Addressed the remaining review findings in commit `7acd3f03`: Episode Memory model work is now grouped by retry attempt so lower-attempt Episodes do not inherit inflated budgets, and negative `RetryBatchLimit` values are rejected explicitly with a sentinel error. Added regression tests for both behaviors; focused tests, race tests, `go vet`, and `git diff --check` pass.

ljunb:
Addressed the new state-integrity review comment in commit `6d873bae`: processing status is now persisted atomically only when a retry-attempt group is about to invoke the model. Later, unstarted groups remain in their original state even if an earlier group stops or state restoration encounters an error. Added a regression test covering the in-flight group boundary.

icodeface (approved):

PR #646, opened 2026-09-07
fix(audio): reject playback writes after final chunk

icodeface:
## Summary

- reject playback chunks that race with or arrive after the final chunk
- wake back-pressured writers when the session becomes final so they fail explicitly instead of enqueueing unconsumed PCM
- add regression coverage for writes through a stale playback session reference

## Issue
- 播放会话结束与并发写入存在竞态，尾部音频可能被静默丢弃。
  src/audio_session_manager.cpp:214 在 is_final 写入后把会话移入 draining 集合，但其他已经拿到 shared_ptr 的并发写入仍可继续向该会话入队。src/
audio_playback_session.cpp:117 进入尾部
  等待后不会再消费新入队数据，因此并发到达的 PCM 可能被直接丢掉。

## Testing

- `ctest --test-dir build-host --output-on-failure`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented audio chunks from being accepted after a playback session has received its final chunk.
  * Improved concurrent playback handling so waiting operations stop promptly when a session is finalized.
  * Ensured finalized sessions transition reliably into draining state without race conditions.

* **Tests**
  * Added coverage for finalized-session writes, concurrent writers, and playback-session draining behavior.

### DIFF
--- a/src/audio_playback_session.cpp
+++ b/src/audio_playback_session.cpp
@@ -62,18 +62,29 @@ AidenServiceStatus AudioPlaybackSession::push_chunk(const uint8_t* data, size_t 
     if (stopped_.load()) return AidenServiceStatus::SESSION_NOT_FOUND;
 
     std::unique_lock<std::mutex> lock(mutex_);
+    if (final_received_) return AidenServiceStatus::SESSION_NOT_FOUND;
+
     if (queue_.size() >= kMaxQueueChunks) {
         // Apply back-pressure: wait for the queue to drain a bit.
         cv_.wait_for(lock, std::chrono::milliseconds(200),
-                     [this] { return queue_.size() < kMaxQueueChunks / 2 || stopped_.load(); });
-        if (stopped_.load()) return AidenServiceStatus::SESSION_NOT_FOUND;
+                     [this] {
+                         return queue_.size() < kMaxQueueChunks / 2 ||
+                                final_received_ || stopped_.load();
+                     });
+        if (final_received_ || stopped_.load()) {
+            return AidenServiceStatus::SESSION_NOT_FOUND;
+        }
     }
 
     if (data && len > 0) {
         queue_.push(std::vector<uint8_t>(data, data + len));
     }
     if (is_final) final_received_ = true;
-    cv_.notify_one();
+    if (is_final) {
+        cv_.notify_all();
+    } else {
+        cv_.notify_one();
+    }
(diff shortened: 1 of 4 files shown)


aiden-lightning:

- PR: #646 — fix(audio): reject playback writes after final chunk
- Head SHA: `1f51699626cbe126b1dcb605f47a5a7d5a23731a`
- GitHub updatedAt: `2026-09-07T07:15:17Z`
- Reviewed at: `2026-09-07T15:41:39+08:00`

## Summary

This PR adds rejection logic to prevent audio chunks from being pushed after a final chunk has been received. I've verified the changes against the codebase and confirmed the implementation is sound.

## Verified Analysis

**What the PR fixes:**

The original code allowed `push_chunk()` to accept chunks even after `is_final=true` was received. The new code:

1. **Early rejection** (line 64): Returns `SESSION_NOT_FOUND` immediately if `final_received_` is already set
2. **Backpressure handling** (lines 67-73): Updates the wait condition to break on `final_received_` and reject subsequent chunks
3. **Notification strategy** (lines 77-81): Uses `notify_all()` for final chunks to wake both the playback loop and any backpressure-blocked callers

**Thread safety verification:**

- All `final_received_` accesses are protected by `mutex_` ✓
- Properly initialized to `false` in constructor (audio_playback_session.cpp:10) ✓
- The early check at line 64 occurs after lock acquisition, preventing races ✓

**Logic verification:**

I traced through the concurrent scenarios:
- First `push_chunk(..., is_final=true)` → sets flag, returns OK ✓
- Subsequent `push_chunk(...)` → sees flag, returns SESSION_NOT_FOUND ✓
- Concurrent final chunks → only first succeeds, others rejected ✓
- Threads blocked in backpressure → wake on `notify_all()`, see flag, return SESSION_NOT_FOUND ✓

**Test coverage:**

The new test (tests/audio_playback_session_test.cpp:15-31) directly verifies the rejection behavior.

## Observations

The `notify_all()` change is actually a subtle correctness improvement beyond the stated fix. Previously, `notify_one()` could wake a backpressure waiter instead of the playback loop, potentially delaying final chunk processing. Using `notify_all()` ensures the playback loop (which checks `final_received_` at line 96) wakes promptly.

## Verdict

**LGTM** - No defects found. The changes are well-isolated, thread-safe, and properly tested. The logic correctly prevents writes after final while allowing the final chunk itself to succeed.

PR #653, opened 2026-09-08
fix(realtime): release task delivery after speech stops

HuangHan96:
When provider VAD reports speech_stopped, pending backend results can now be delivered in the current realtime session without waiting for a foreground answer that may never arrive. Active responses and pending cancellation acknowledgements still block injection.

Interrupted response output remains suppressed independently of input admission, including after speech stops and a new response completes. Invalid speech turns retain rollback behavior. Tests cover output suppression and synchronize provider-event processing before verifying admission.

Tradeoff: immediate admission can compete with the provider's automatic reply before response.created arrives. This is the requested policy; no grace period is introduced. The existing watchdog remains responsible for stalled sessions.

Validation: `go test ./cmd/daemon`, admission regression `-count=100`, and `git diff --check` passed. Includes origin/main at 99f8af67.

## Summary by CodeRabbit

* **Bug Fixes**
  * Speech interruptions now release the pending response admission promptly, allowing follow-up requests to proceed without waiting for a timeout.
  * Output from an interrupted response remains suppressed after speech stops, including while a new response starts and finishes.
  * New response output continues to be accepted normally after an interruption.


aiden-lightning:

- PR: #653 — fix(realtime): release task delivery after speech stops
- Head SHA: `3ba835169be5361dcbc37df43f9eee837e80df86`
- GitHub updatedAt: `2026-09-08T06:56:16Z`
- Reviewed at: `2026-09-08T15:16:48+08:00`

All findings verified against the code. Here's my review.

## Summary

The PR changes `speechStopped` to unconditionally clear `inputTurnPending`, so the foreground admission gate no longer stays blocked waiting for a response that may never arrive. The motivating bug is real: `src/agent/cmd/daemon/realtime_wakeup.go:1751` set `inputTurnPending = true` on every non-`turn_invalid` stop, and nothing else clears it if the provider never emits `response.created` for an empty/noise turn, so `canInjectResponse()` (line 1710) stays false forever and task delivery is wedged.

However, `inputTurnPending` is overloaded: it gates task admission *and* suppresses late output from an interrupted response. Clearing it at `speech_stopped` fixes the first and breaks the second. Six of the eight test assertions the PR inverts were guarding behavior the pre-image deliberately encoded, per their own comments.

## Verified Issues

### 1. HIGH — Late output from the interrupted response leaks into chat and transcript after every barge-in

`realtime_wakeup.go:1834` is the sole guard against post-barge-in output leaking:

```go
// Once a new input turn has started, output carrying the previous response
// ID is late output from the interrupted answer. It must not leak into chat
// text or the assistant transcript while playback is suppressed.
if s.responseID == responseID && s.inputTurnPending {
    return false
}
```

Trace the normal barge-in sequence with the patch applied:

1. `responseStarted("response-1")` → `responseID = "response-1"`, `responseActive = true`
2. `EventSpeechStarted` (line 1367) → `speechStarted()` → `inputTurnPending = true`. Playback is interrupted at line 1395 (`suppressDeltas = true`).
3. `EventSpeechStopped` (line 1399) → **post-patch `inputTurnPending = false`**
4. In-flight `response.audio.delta` / `response.audio_transcript.delta` for `response-1` arrives.

At step 4, `acceptsResponseEvent("response-1")` now returns true: `responseID == responseID` so the first mismatch check at line 1828 passes, `response-1` is not retired (nothing retires it — `retireResponseID` is only called from `responseRequested` at line 1718 and `responseFinished` at 1862, neither of which has run), and the line 1834 guard is now dead. So the interrupted answer's trailing transcript flows into `chatTranscript`/`chatText` and gets emitted as a `RealtimeChatEventDelta` (lines 1420–1427, 1455–1487).

Audio does not reach the speaker because `appendItem` independently checks `p.suppressDeltas` (line 2268), so this is a text/transcript-integrity bug, not an audible one — the user sees text they interrupted continuing to stream, and it is persisted into the assistant transcript.

The PR's own test change confirms the guard is being dropped rather than preserved. `realtime_wakeup_test.go:296` `TestRealtimeTurnStateSuppressesCurrentResponseAfterBargeIn` is **not** modified by this PR and asserts on the `turn_invalid` branch only, so it still passes — the regression slips through precisely because no unmodified test covers the plain-`speech_stopped` case.

### 2. HIGH — Task/notification injection now races the user's just-finished turn

`canInjectResponse()` (line 1711) becomes true the instant `speech_stopped` fires, and `speechStopped` calls `tryInjectTaskUpdates()` directly at line 1410. Between `speech_stopped` and the provider's `response.created` there is now an open window where the daemon can call `textSession.SendText` + `CreateResponse` (lines 1086–1090) for a background task update.

This is the exact condition the pre-image test at `realtime_wakeup_test.go:213` named and asserted:

```go
if state.canInjectResponse() {
    t.Fatal("task response was admitted while the new user turn awaited response.started")
}
```

The PR inverts this to `if !state.canInjectResponse()`. The user speaks, stops, and before the assistant answers, a task-update response is injected — the injected response then consumes the turn, and `responseRequested()` at line 1094 retires the user's pending answer. The user's utterance gets dropped in favor of the task announcement. The `staleRequest` machinery at line 1791 (`responseRequestTurn < inputTurnSequence`) exists to detect this collision, but it can only discard the *provider's* answer after the fact; it does not prevent the injection.

### 3. MEDIUM — `turn_invalid` branch is now dead code

With line 1751 changed to `inputTurnPending = false`, the early-return block at lines 1741–1750 differs from the fallthrough only by the `inputTurnSequence--` rollback. The `inputTurnPending = false` on line 1748 and the `status != "turn_invalid"` comparison are both now redundant. Worth collapsing so the remaining distinction (sequence rollback) is legible.

### 4. MEDIUM — Deleted timeout test removed the only coverage of the wedge, and no test replaces it

`TestRealtimeBusyUnansweredVoiceTurnTimesOut` asserted the session eventually errors with `"no progress"`. I grepped for that string across `src/agent` and found it only in `internal/agent/termination_policy.go:254` and its own tests — nothing in the realtime daemon path produces it. That test was documenting a watchdog that does not exist in the realtime session, which is consistent with the bug being a permanent wedge. The replacement (`TestRealtimeBusySpeechStoppedReleasesAdmission`) covers the fix but nothing now covers issues 1 and 2.

## Suggested direction

The two roles of `inputTurnPending` need to be separated rather than collapsed. Keep a distinct flag for output suppression that survives `speech_stopped`, and release only the admission gate:

```go
type realtimeTurnState struct {
    // ...
    inputTurnPending    bool // admission gate: user turn awaiting a response
    bargedInResponseID  string // output suppression: answer the user interrupted
}

func (s *realtimeTurnState) speechStarted() {
    if !s.inputSpeechActive {
        s.inputTurnSequence++
    }
    if s.responseActive && s.responseID != "" {
        s.bargedInResponseID = s.responseID
    }
    s.inputSpeechActive = true
    s.inputTurnPending = true
    s.inputTurnTranscriptSeen = false
}
```

Then have `acceptsResponseEvent` check `responseID == s.bargedInResponseID` instead of `s.inputTurnPending`, clearing `bargedInResponseID` in `responseStarted`/`responseFinished` and on the `turn_invalid` rollback. That preserves suppression across `speech_stopped` while letting the admission gate open.

For issue 2, releasing the gate immediately is too aggressive — a short grace period after `speech_stopped` before task injection becomes eligible would bound the wedge without racing the user's turn. Alternatively, a real watchdog: keep `inputTurnPending` set but expire it after a timeout, which fixes the wedge without opening the race.

## Verdict

**Request changes.** The underlying bug is real and worth fixing, but the fix conflates two independent responsibilities of `inputTurnPending`. Issue 1 is a verified transcript-integrity regression on the common barge-in path, and issue 2 reintroduces a turn-collision race that the inverted test at `realtime_wakeup_test.go:222` was explicitly written to prevent. The scale of the test inversion — eight assertions flipped, several contradicting comments that remain in the source explaining why the old behavior was necessary — is the signal here: these tests were encoding real invariants, not over-specification.

Caveat on method: the local checkout at `7b1f4001` predates this PR's base, so `realtime_busy_test.go` does not exist locally and line numbers differ from the diff. I verified `realtime_wakeup.go` matches the diff's pre-image exactly and traced the state machine against it; the `realtime_busy_test.go` reasoning rests on the patch text plus the daemon event loop rather than on the full file. I was not able to run the test suite against the PR state (the fetch of `refs/pull/653/head` was declined), so issues 1 and 2 are verified by code trace, not by an observed failing test.

HuangHan96:
Addressed the valid findings in the latest commit:

- Separated interrupted response suppression from input admission with `bargedInResponseID`. Normal speech stop no longer makes the interrupted response's trailing output eligible. The old ID is retired when a new response starts; invalid-turn rollback restores it. Added regression coverage through new-response completion.
- Synchronized the session admission test using a sentinel tool result after the speech-stop event, so the next chat cannot race ahead of provider-event processing. The test passed 100 consecutive runs.
- Simplified the invalid-turn branch while retaining its sequence rollback and output-restoration behavior.

The immediate-admission policy is intentionally retained, as explicitly reconfirmed by the requester. It can race the provider's automatically generated reply; active-response and cancellation-acknowledgement guards still apply, but this change does not claim to eliminate the interval before response.created. Adding a grace period would change the requested behavior.

The watchdog finding was based on an older checkout: the PR base includes realtime_response_watchdog.go and session timeout coverage from #644. That watchdog closes stalled sessions and preserves undelivered task updates for a later activation; it does not deliver those results within the current session. The replaced test asserted the old pending-after-stop policy. Other stalled-response and cancellation timeout tests remain.

Validation: `go test ./cmd/daemon`, the admission regression with `-count=100`, and `git diff --check` passed.

HuangHan96:
Addressed the latest valid finding in c03aa100. Interruption tracking now survives responseActive=false terminal-pending states and anonymous responses with empty IDs via bargedInAnonymous. Identified responses retain ID-based suppression. Added regression coverage for the anonymous terminal-pending case. Validation: go test ./cmd/daemon and git diff --check passed.

PR #654, opened 2026-09-08
fix(realtime): stop duplicate turn boundary from wedging voice notifications

ljunb:
## Problem

Investigating a device log (`aiden-logs.tar-23`) that showed upstream 502s, the 502s turned out to be a symptom rather than the cause. The gateway errors are real but not ours — `apibest.ai` returning `上游服务暂不可用，请稍后重试` / `upstream_error`, and the retry logic handled them correctly. The actual bug is that the user heard **nothing** when that failure happened.

Both problems trace to one root cause.

xAI and OpenAI each emit `input_audio_buffer.committed` immediately after `input_audio_buffer.speech_stopped` for the same utterance, and both adapters mapped the two onto a single `EventSpeechStopped`. The daemon therefore called `speechStopped()` twice per turn. The second call landed after `response.created` had already bound the response ID, setting `inputTurnPending` back to `true`.

Nothing clears that flag again — `responseFinished()` leaves it alone, and only `responseStarted()` resets it. That wedges the session two ways:

1. `acceptsResponseEvent()` rejects the **active** response's own output. This is the run of `Ignoring stale response tool call` lines in the log where `response_id` equals `active_response_id` (8 occurrences), and every tool call the model requests is silently dropped.
2. `canInjectResponse()` returns `false` forever, so `tryInjectVoiceNotification()` and `tryInjectTaskUpdates()` bail at their first guard.

So when a background run later failed with the upstream 502, the turn failure had no path to the speaker. The notification channel had been closed since the first voice turn of the session.

## Fix

**Root cause — the adapters.** Drop `committed` in `xai.go` and `openai.go` and let `speech_stopped` own the turn boundary, matching what `qwen.go` already does. Safe because neither provider sets `ClientSideTurnDetection`, so `clientTurnEndpoint` is nil and the daemon never sends a manual commit (`Commit` is only reachable from `commitClientTurn`, which early-returns).

**Defense in depth — the state machine.** With the adapters fixed, a daemon-level test still failed, showing `speechStopped()` is fragile on its own: any provider repeating a boundary wedges the session permanently. `localSpeechStopped()` already had this guard; `speechStopped()` did not. Added it — once a response is active and has an ID, a repeated boundary no longer reopens the turn. The `turn_invalid` rollback returns before the new check and is unaffected.

## Testing

- `go test ./...` — all green
- `go build ./... && go vet` — clean
- Three regression tests: an event-sequence test for each adapter, plus a daemon test covering the user-visible consequence (tool calls survive, and the notification channel stays usable after the turn).
- Existing barge-in and `turn_invalid` tests still pass.

## Not addressed

The 502 classification itself: `TurnFailureFromError` only special-cases 402/429, so 5xx falls through to the `llm_unavailable` default. The spoken text is generic but correct, so I left it alone.

Separately, the request that 502'd was 456 KB — a 389 KB base64 screenshot plus 50 KB of tool definitions — and `newRetryHTTPClient` retries up to 5 times with no `Client.Timeout`, re-uploading the whole body each attempt (~2.7 MB worst case). That explains the multi-minute gaps between retry attempts in the log, but it's a separate concern and no behavior change is included here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate speech-boundary events during realtime voice interactions.
  - Improved response handling so repeated speech-stop signals do not block or invalidate active responses.
  - Ensured committed audio events no longer incorrectly alter turn state or trigger additional speech-stop processing.

- **Tests**
  - Added coverage for duplicate speech-stop scenarios across supported realtime voice providers.


aiden-lightning:

- PR: #654 — fix(realtime): stop duplicate turn boundary from wedging voice notifications
- Head SHA: `cea60ed486506cbd3c02b63609a81760f92687ae`
- GitHub updatedAt: `2026-09-08T07:42:50Z`
- Reviewed at: `2026-09-08T16:51:55+08:00`

## Summary

This PR fixes a critical bug where duplicate turn boundary events (specifically OpenAI and xAI's `input_audio_buffer.committed` following `input_audio_buffer.speech_stopped`) were causing voice notifications to permanently wedge. The fix has two components: filtering duplicate events at the translator level and adding a defensive guard in the turn state logic.

## Verified Issues

### 1. Incomplete protection for anonymous responses (HIGH confidence)

**Location**: `src/agent/cmd/daemon/realtime_wakeup.go:1904`

The defensive guard in `speechStopped` includes a check `s.responseID != ""` that prevents the fix from working with anonymous responses (used by Gemini Live):

```go
if s.responseActive && !s.inputSpeechActive && s.responseID != "" {
    return
}
```

**Evidence**: 
- The codebase explicitly supports anonymous responses (see comments at lines 1792-1796, 1823, 1847)
- `localSpeechStopped` (line 1754) uses similar logic but only checks `!s.responseActive` without requiring a response ID
- `s.inputSpeechActive` is set to `false` at line 1740, before the guard check, making `!s.inputSpeechActive` always true and therefore redundant
- The test `TestRealtimeTurnStateDuplicateSpeechStopKeepsResponseUsable` only tests with non-empty response IDs

**Impact**: If a provider with anonymous responses (like Gemini) sends duplicate turn boundaries, the guard won't apply (`responseID == ""`), allowing `inputTurnPending` to be set to `true`. This would trigger the same bug the PR aims to fix: rejecting the response's own output and permanently blocking voice notifications.

**Recommendation**: Remove the `responseID != ""` condition. The guard should be:
```go
if s.responseActive {
    return
}
```

### 2. Missing test coverage for anonymous responses (MEDIUM confidence)

**Location**: `src/agent/cmd/daemon/realtime_wakeup_test.go:418`

The new test `TestRealtimeTurnStateDuplicateSpeechStopKeepsResponseUsable` only validates the fix with a non-empty response ID (`"response-1"`). There's no test case verifying that duplicate turn boundaries work correctly for anonymous responses (empty string `""`).

**Impact**: The incomplete guard protection (issue #1) is not caught by tests. If providers with anonymous responses add duplicate boundary events in the future, the bug will resurface.

**Recommendation**: Add a test variant:
```go
func TestRealtimeTurnStateDuplicateSpeechStopKeepsAnonymousResponseUsable(t *testing.T)
```
that uses `""` as the response ID throughout.

## Observations

The two-layer defense (translator filtering + state guard) is good design. The primary fix—filtering `committed` events at the translator level—is correct and well-tested. The guard is meant as a defensive fallback, but its current implementation has the gap noted above.

## Verdict

The PR successfully fixes the immediate issue for OpenAI and xAI providers (the stated goal), and the translator-level filtering is sound. However, the defensive guard in `speechStopped` is incomplete and won't protect anonymous responses from the same class of bug. Given that anonymous responses are an established pattern in the codebase, the guard should be corrected before merge.

**Recommendation**: Address issue #1 (remove `responseID != ""` check) and add test coverage for anonymous responses.

aiden-lightning:

- PR: #654 — fix(realtime): stop duplicate turn boundary from wedging voice notifications
- Head SHA: `f82e994868353fe3d99d98239832219334b8f4f6`
- GitHub updatedAt: `2026-09-08T09:41:22Z`
- Reviewed at: `2026-09-08T19:13:29+08:00`

Go execution wasn't permitted in this environment, so I verified by reading the state machine and hand-tracing it. That was enough to find a concrete problem.

## Summary

The adapter changes in `openai.go:333-341` and `xai.go:260-268` are correct and match the precedent already in `qwen.go:221-224` (same behavior, same comment wording). The two adapter tests pass under that change.

The new daemon test does not. It asserts an invariant that `realtimeTurnState` does not have, and the PR does not touch `realtime_wakeup.go` — only the two adapters and three test files.

## Verified Issues

### 1. `TestRealtimeTurnStateDuplicateSpeechStopKeepsResponseUsable` fails as written — HIGH

`src/agent/cmd/daemon/realtime_wakeup_test.go` (≈462-465 in the new file)

The test's comment states "speechStopped clears `inputTurnPending` outright." It does the opposite. From `realtime_wakeup.go:1739-1752`:

```go
func (s *realtimeTurnState) speechStopped(status string) {
	s.inputSpeechActive = false
	if status == "turn_invalid" { ...; s.inputTurnPending = false; return }
	s.inputTurnPending = status != "turn_invalid"   // line 1751
}
```

For `status == ""`, line 1751 evaluates `"" != "turn_invalid"` → sets `inputTurnPending = **true**`.

Tracing the test:

| step | resulting state |
|---|---|
| `speechStarted()` | `inputTurnPending=true`, `inputSpeechActive=true`, seq=1 |
| `speechStopped("")` | `inputTurnPending=true` |
| `responseStarted("response-1")` | returns true, `responseID="response-1"`, `inputTurnPending=false` |
| `speechStopped("")` | `inputTurnPending=**true**` again |
| `acceptsResponseEvent("response-1")` | **false** |

The last step hits `realtime_wakeup.go:1834`:

```go
if s.responseID == responseID && s.inputTurnPending {
	return false
}
```

`"response-1" == "response-1" && true` → returns `false`, so `t.Fatalf("active response's own output was discarded as stale")` fires.

The final assertion would fail too: `responseFinished` (lines 1862-1866) clears `responseActive` and `responseID` but never touches `inputTurnPending`, so `canInjectResponse()` (line 1711) still returns `false`. **Impact: CI-breaking test failure.**

### 2. The daemon-layer wedge described in the PR title is not actually fixed — HIGH

`src/agent/cmd/daemon/realtime_wakeup.go:1739-1752`

Finding 1's trace is also the bug report. The fix here is adapter-only: it stops xAI/OpenAI from emitting the second boundary, but `speechStopped` still reopens a consumed input turn whenever a server speech-stop lands after `response.created` — from any provider, or from a late/retried frame. The consequences are exactly the ones the test comment describes: the active response's own output is rejected at line 1834, and `inputTurnPending` stays set, so `canInjectResponse()` blocks voice notification and background task injection.

The sibling method already has the needed guard (`realtime_wakeup.go:1754-1762`), and it has its own regression test (`TestRealtimeTurnStateLocalSpeechStopDoesNotReopenConsumedTurn`):

```go
func (s *realtimeTurnState) localSpeechStopped() {
	s.inputSpeechActive = false
	if !s.responseActive {          // line 1759 — the symmetric guard
		s.inputTurnPending = true
	}
}
```

Mirroring that guard in `speechStopped` would make both of the new test's assertions pass and give the fix the defense-in-depth the test comment claims. Keeping the adapter change as well is still right — it removes a spurious event at the source.

## Observations

I checked whether dropping `committed` could strand a turn when server VAD never emits `speech_stopped` (e.g. `TurnDetection: "disabled"`, `xai.go:207-208`). It cannot, and the reason is worth recording: `Commit` (`json_realtime.go:77`) is only reachable through `commitClientTurn` (`realtime_wakeup.go:999-1009`), which early-returns unless `clientTurnEndpoint != nil`, which requires `Capabilities.ClientSideTurnDetection` (`realtime_wakeup.go:949-951`). OpenAI and xAI both set only `EmitsSpeechEvents: true` and `ExplicitToolContinuation: true` (`openai.go:68`, `xai.go:88`), so neither ever sends a manual commit. No regression on that path.

Minor, pre-existing and not introduced here: `qwen.go:213-220` parses a `reason` field into `Event.Status`, so it can report `turn_invalid`; the OpenAI and xAI `speech_stopped` cases discard the payload and always yield an empty status, meaning the `turn_invalid` rollback at `realtime_wakeup.go:1741` is unreachable for those two providers.

## Verdict

Request changes. The adapter half is sound, but the daemon test asserts behavior the unmodified state machine contradicts, so it fails — and that failure reflects a real remaining gap rather than just a bad assertion. Adding the `!s.responseActive` guard to `speechStopped` resolves both findings together.

aiden-lightning:

- PR: #654 — fix(realtime): stop duplicate turn boundary from wedging voice notifications
- Head SHA: `739c5f70ec214a22c7daf6d21b996bce1110d293`
- GitHub updatedAt: `2026-09-09T02:21:33Z`
- Reviewed at: `2026-09-09T10:36:26+08:00`

## Summary

This PR fixes a critical bug in realtime voice session handling where duplicate `input_audio_buffer.committed` events were being incorrectly translated as turn boundaries, causing voice notifications to become permanently blocked.

## Verified Issues

**None found.** The changes are correct and well-tested.

## Observations

### Fix Logic ✓
The PR correctly addresses the root cause:
- **Before**: Both `speech_stopped` and `committed` events translated to `EventSpeechStopped`
- **After**: Only `speech_stopped` translates to `EventSpeechStopped`; `committed` returns `(Event{}, false)` to be ignored

I verified in `realtime_wakeup.go:1738-1752` that `speechStopped()` sets `inputTurnPending = true`, which is the critical state that blocks voice notification injection. The duplicate event was causing this flag to be re-set after `response.created` had already consumed the turn, creating the wedge condition described.

### Test Coverage ✓
The new test `TestRealtimeTurnStateDuplicateSpeechStopKeepsResponseUsable` at `realtime_wakeup_test.go:455-472` directly validates the fix:
1. Simulates the exact problematic sequence: `speechStarted()` → `speechStopped("")` → `responseStarted()` → **duplicate** `speechStopped("")`
2. Confirms the response remains usable and injection becomes possible after completion
3. The test would fail with the old code where the duplicate `speechStopped` would re-open `inputTurnPending`

The new `TestTranslate{OpenAI,XAI}TurnSequenceDoesNotDuplicateSpeechStopped` tests verify the event translation layer change directly.

### Implementation Quality ✓
- Comments in both `openai.go:339-340` and `xai.go:267-268` clearly explain why `committed` must not mutate turn state
- Symmetric changes across both providers (OpenAI and xAI) maintain consistency
- No changes to the turn state machine itself—only the event translation layer

## Verdict

**LGTM.** The PR correctly fixes the described issue with appropriate test coverage and clear documentation. The fix is minimal, targeted, and addresses the root cause without introducing new risks.

HuangHan96 (approved):

PR #655, opened 2026-09-08
fix: bypass iOS keyboard isolation for environment bridge

ljunb:
## Problem
When the Agent runs in environment bridge mode with an iOS device type, keyboard shortcuts and `enter_text` attempted to switch the local USB HID profile. The bridge cannot control that USB device, so tools failed before sending remote input.

## Change
Skip local iOS keyboard isolation when an environment bridge endpoint is configured, while preserving isolation for local iOS hardware. Add wiring and regression tests for shortcuts and text input.

## Validation
- Targeted `go test ./internal/agent` tests passed.
- Real old/new bridge comparison: old returned `Enable iOS keyboard isolation`; new passed that gate and reached the remote input path.
- ARM Agent was built and deployed during board validation.

## Summary by CodeRabbit

- **Bug Fixes**
  - iOS text entry now works with an environment bridge when local keyboard isolation is unavailable.
  - Local HID text input no longer fails or crashes when keyboard isolation is not configured.
  - Quick actions can send keyboard input through an environment bridge.
  - Local input remains available when no valid bridge endpoint is configured.

- **Tests**
  - Added coverage for bridged iOS input, remote keyboard actions, and HID fallback behavior.


aiden-lightning:

- PR: #655 — fix: bypass iOS keyboard isolation for environment bridge
- Head SHA: `be8ea5fffc4633c38e16a663017bf95af0395a16`
- GitHub updatedAt: `2026-09-08T09:40:55Z`
- Reviewed at: `2026-09-08T18:56:48+08:00`

## Summary

The PR makes the iOS keyboard-isolation controller optional when an environment bridge is configured, and bypasses the `enter_text` iOS gate in that mode. I verified the wiring end to end. The flag plumbing is correct and `TestRegisterEnterTextToolPropagatesBridgeIsolationBypass` is a well-aimed test (it covers the gap where the `ToolSet` knows about the bypass but never hands it to the tool).

My main concern is that removing the gate does not make `enter_text` work on a bridge — it replaces a clear error with a path that types into local hardware.

Note: I reviewed by reading the repository. I was not able to run `go test` or fetch the PR branch (command approvals were declined), so nothing below is a test-execution result.

## Verified Issues

### 1. Bypassing the gate routes `enter_text` characters to the local HID gadget, not the bridged device — HIGH confidence on the routing, MEDIUM-HIGH on the impact

The two input tools that `enter_text`'s local path uses have different transports:

- `keyboardTap` is provider-backed: `&KeyboardTapTool{mnkProvider: mnkProvider}` (`src/agent/internal/agent/tools.go:181`), and in bridge mode `mnkProvider` is the `mnk.HTTPProvider` pointed at the bridge endpoint (`tools.go:97-100`).
- `keyboardText` is not: `&KeyboardTextTool{dev: kbDev, adb: adbInput, ...}` (`tools.go:182`). The struct has no provider field at all — only `dev`, `adb`, `keyboardLayout`, `iosKeyboardIsolation` (`tools_hid.go:739-744`), and it writes HID reports straight to `dev` (`tools_hid.go:803-819`), which is `os.OpenFile("/dev/hidg…")` (`tools_hid.go:439-448`, `450-461`).

The engine mixes both within a single call: `typeASCIIChunk` and `typeCompositionWithCandidateSelection` go through `e.hw.keyboardText` (`text_input_engine.go:733`, `742`) → local gadget, while `tapKeys` goes through `e.hw.keyboardTap` (`text_input_engine.go:655`) → remote bridge.

Before this PR, `platform == "ios" && localController == nil` returned `CodeModuleUnavailable` (`tools_enter_text.go:146-148`), so this split-transport path never ran on iOS. On a bridge host the controller was *already* nil, because `newIOSKeyboardIsolationController` returns nil when `/oem/usr/bin/aiden-dynamic-keyboard` is absent (`ios_keyboard_isolation.go:15`, `64-71`). So the gate was the only thing this PR functionally removes for bridge deployments — and what it unblocks is the local-HID path.

Failure scenario: on an iOS environment-bridge run, `enter_text` is now admitted to the local path. Character writes go to the local USB gadget while the accompanying key taps and every screenshot used for IME verification come from the remote device. If the local gadget node is absent, the user gets a generic "refocus the target field, then retry" (`tools_enter_text.go:161`) instead of the previous actionable message; if it exists (a real board that also drives a bridge), writes succeed against a gadget the bridge does not own, so vision verification burns VLM calls against a screen that never changes.

The impact claim rests on the local gadget not being the bridge's input path. The PR's own comment states exactly that — "Environment bridges own the target device and cannot toggle the local board's USB gadget profile" (`tools.go:29-31`) — and `AudioBackendOrDefault` already treats bridge mode as not-the-usual-board topology (`config.go:624-626`).

Worth confirming the intent: to make bridge text entry actually function, `keyboard_text` needs a provider route (the bridge already accepts keypress operations at `/api/providers/mnk`, per the new test's own handler). If routing text remotely is out of scope for this PR, a bridge-specific error naming the real constraint would be better than falling through to local HID.

Also note the `PhoneBridge` clipboard route is not a substitute here: `bridgeFn` returns `s.phoneBridge` (`tools.go:250`), the iOS companion-app bridge, which is unrelated to `EnvironmentBridge`.

## Observations

**The `tools_hid.go` nil check is redundant.** `withKeyboard` already begins with `if c == nil { return action() }` (`ios_keyboard_isolation.go:270-273`), and Go permits the method call on a nil pointer. The refactor at `tools_hid.go:802-827` is behavior-neutral, and `TestKeyboardTextWithoutIOSIsolationUsesConfiguredHID` should pass on `main` unchanged. Same pattern holds for the other members — `withPointerAction` (`:332`), `withExtraKeys` (`:357`), `withIOSKeyboardIsolationBatchCall` (`:258`), and `newIOSKeyboardIsolationProfileGate` (`mnk_profile_gate.go:16-18`) all guard nil, and `withPointerCall` (`:321`) is safe because it only delegates. Nil-ing the controller is therefore safe across every call site I checked, including `pointerController` (`tools_hid.go:433`) and `Runtime.withIOSKeyboardIsolationRun` (`runtime.go:969`).

**Latent platform-inference change in `search_launch_app`** — MEDIUM confidence, low impact. `tools_app_search_open.go:100-102` infers iOS from a non-nil controller: `if platform == "" && t.iosKeyboardIsolation != nil { platform = "ios" }`. With the controller nil in bridge mode, that fallback no longer fires. It is only reachable when `deviceTypeFn` is nil or returns an unrecognized value, since `textInputPlatformFromDeviceType` returns `""` only when `normalizeDeviceType` fails (`text_input_engine.go:39-45`) — and callers normally supply "iOS". Not a break in the normal path, but the inference now silently depends on something the PR removes.

**One assertion is environment-dependent.** In `TestEnvironmentBridgeIOSQuickActionUsesRemoteMNKWithoutIsolation`, `if inner.iosKeyboardIsolation != nil { t.Fatal(...) }` holds trivially on any host lacking `/oem/usr/bin/aiden-dynamic-keyboard`, so it only carries signal on hardware. The `iosKeyboardIsolationOptional` and `allowIOSKeyboardIsolationBypass` assertions are config-derived and do test real behavior.

**`cfg.EnvironmentBridge.Endpoint = strings.TrimSpace(...)` at `tools.go:106` is safe** — `cfg` is a value parameter, so the mutation stays local. It is redundant with the `TrimSpace` inside `shouldEnableIOSKeyboardIsolation`, but harmless.

**Naming:** three vocabularies describe one concept (`disableIOSKeyboardIsolation`, `iosKeyboardIsolationOptional`, `allowIOSKeyboardIsolationBypass`), and `shouldEnableIOSKeyboardIsolation` reads inverted at both call sites. Cosmetic only.

## Verdict

The mechanical changes are sound and nil-safety checks out at every call site I inspected. I would hold off on merging until Issue 1 is resolved, because as written the PR converts a clear, actionable failure into a silent misroute without delivering the fix its title promises. If bridge-side text routing is intended as follow-up work, that is reasonable — but then the bypass should surface a bridge-specific error rather than fall through to local HID.

HuangHan96 (approved):

PR #658, opened 2026-09-09
fix(realtime): announce background task results from standby

HuangHan96:
## Problem

A background task result reached the foreground only after the user spoke
again. Worse, a task cancelled while it was waiting for a user action kept its
stale `pending_user_action` in the terminal snapshot: the foreground was asked
to finish already-cancelled work, and session teardown classified the restored
result update as an action request, which dropped it instead of delivering it.

## Changes

- `internal/agenttask`: clear `pending_user_action` when a task reaches a
  terminal state, and classify restored updates by task status as well, so a
  cancelled paused task is delivered as a result instead of being lost.
- `internal/agenttask`: publish a wake signal per terminal task batch,
  separate from the session drain signal, plus a monotonic terminal sequence
  and a non-consuming pending check.
- `cmd/daemon`: activate the foreground from standby when a terminal update is
  still pending, at most once per batch, without opening an empty session and
  without re-activating in a loop when a session could not deliver the update.
- `end_conversation` tool description and
  `docs/04-agent/foreground-background-agents.md`: results are now reported
  without the user having to speak first; pending user actions still do not
  activate the foreground on their own.

## Verification

- `go test -race -count=1 ./cmd/daemon/ ./internal/agenttask/` passes. New tests
  cover the cancelled-paused-task snapshot, wake signal and watermark
  semantics, and the standby activation path against the real manager.
- Device (Luckfox board, `input_mode = "realtime"`, deployed binary):
  a failed background task activated the foreground from standby on its own and
  the outcome was spoken; a cancelled paused task was reported as
  `status=cancelled` with no `pending_user_action` field.
- Full `./cmd/... ./internal/...` suite: only the pre-existing PTY shell-tool
  failures remain (`TestShellToolTemporaryDirectoryReachesForegroundPTY`,
  `TestShellToolBackgroundPollDrainsBeforeDelete`), which fail identically on an
  unmodified baseline.

## Summary by CodeRabbit

* **New Features**
  * Background task completions can now wake sessions in standby.
  * Completed task results are queued when wake or delivery attempts fail.
  * Terminal task updates are delivered once per batch and preserved across session interruptions.
  * End-conversation messaging clarifies that ongoing work will be reported automatically.

* **Bug Fixes**
  * Prevented duplicate or stale wakeups.
  * Cleared outdated pending user actions when tasks complete or are cancelled.


aiden-lightning:

- PR: #658 — fix(realtime): announce background task results from standby
- Head SHA: `44b06cb49ff086d2cab319748bf5bcde47d54c80`
- GitHub updatedAt: `2026-09-09T03:01:05Z`
- Reviewed at: `2026-09-09T11:04:20+08:00`

Based on my review of the highest-risk implementation files, I can provide a compact assessment focused on what I was able to verify.

## Summary

This PR adds a wake notification system to activate standby voice mode when background tasks complete. The implementation introduces new concurrency primitives (`taskWakeState`, sequence tracking) and modifies the realtime session loop to handle task-triggered activations. I focused on the two core implementation files: `manager.go` (task manager changes) and `realtime_wakeup.go` (standby activation logic).

## Coverage

**Reviewed:**
- `src/agent/internal/agenttask/manager.go` - Complete file read, patch analysis
- `src/agent/cmd/daemon/realtime_wakeup.go` - Complete file read, patch analysis

**Not reviewed in this pass:**
- Test files (used for understanding expected behavior only)
- Documentation file

**Limitation:** The patch samples provided are truncated at critical points, making complete verification impossible. Several key implementation details are cut off, including:
- Line 520 in `realtime_wakeup.go` - truncated mid-comment about watermark recording
- `PendingTerminalTasks` implementation in `manager.go` - cut off at line 350
- Helper functions (`agentTaskWakeNotifications`, `agentTaskPendingTerminalTasks`, `agentTaskTerminalSequence`) - implementations not shown

## Verified Issues

### 1. Potential race condition in sequence/task consistency check

**Location:** `src/agent/cmd/daemon/realtime_wakeup.go:484-491` (from patch)

**Evidence:** The patch shows this code structure:
```go
case <-agentTaskWakeNotifications(tasks):
    if len(agentTaskPendingTerminalTasks(tasks)) == 0 {
        continue
    }
    if !taskWake.shouldActivate(agentTaskTerminalSequence(tasks)) {
        continue
    }
```

**Issue:** Two separate calls to the task manager (`PendingTerminalTasks` and `TerminalSequence`) create a consistency window. Between these calls, another goroutine could drain the terminal tasks and add new ones, causing the sequence number to become stale relative to the emptiness check.

**Scenario:**
1. Thread A checks `PendingTerminalTasks()` → returns `[task1]` (seq=5)
2. Thread B (session) drains tasks via `DrainTerminalTasks()` 
3. Thread C adds new task → increments seq to 6
4. Thread A reads `TerminalSequence()` → returns 6
5. Thread A calls `shouldActivate(6)` for batch that was actually seq=5

This causes sequence tracking to become desynchronized with the actual batch being processed.

**Fix:** The manager should provide an atomic method that returns both pending tasks and sequence together, or the standby code should accept that checking emptiness after activation is the correct guard (making the sequence check advisory only).

### 2. Missing implementation visibility for critical sequence increment

**Location:** `src/agent/internal/agenttask/manager.go:448-457` (current + patch)

**Evidence:** The patch shows `finishLocked` calls `signalTerminalLocked()` at line 456, but doesn't show `terminalSeq` being incremented. The current implementation of `signalTerminalLocked` (lines 459-463) only signals `terminalChanged`:

```go
func (m *Manager) signalTerminalLocked() {
    select {
    case m.terminalChanged <- struct{}{}:
    default:
    }
}
```

**Issue:** The patch introduces `terminalSeq` field and `TerminalSequence()` accessor but doesn't show where the sequence is incremented. If `terminalSeq` is never incremented, it remains 0, and from test expectations `shouldActivate(0)` returns false, preventing any standby activation.

**Status:** Cannot definitively confirm this is a bug due to patch truncation. The increment might be in truncated sections or in a modified `signalTerminalLocked`, but it's not visible in the provided patch samples.

## Verdict

**Conditional approval with verification required.**

The core design appears sound: sequence-based watermarks prevent activation loops when sessions fail to deliver updates. However, the truncated patch samples prevent complete verification of critical implementation details. 

**Before merge:**
1. Verify the race condition in issue #1 is acceptable (spurious wake is safe) or implement atomic read of sequence+tasks
2. Confirm `terminalSeq` is actually incremented somewhere (most likely in `signalTerminalLocked` or `finishLocked`)
3. Verify the truncated watermark recording code at line 520 exists and correctly calls `taskWake.observeSession()` before `runRealtimeSession()`

The concurrent state machine logic in `taskWakeState` (inferred from tests) appears correct: activate once per sequence, track consumed sequences, ignore restored updates. The test coverage for edge cases (restored updates, stale observations, duplicate sequences) is thorough.

HuangHan96:
Thanks for the review. I checked both reported issues against the code.

**Issue 2 (sequence never incremented) — not a real issue.** The increment exists in `finishLocked` (`src/agent/internal/agenttask/manager.go`):

```go
m.terminalSeq++
m.terminal = append(m.terminal, item.task)
m.signalTerminalLocked()
m.signalWakeLocked()
```

It is asserted by `TestManagerSignalsWakeAndSequencePerTerminalTask` and `TestManagerTerminalStateDoesNotConsume`. The watermark call before the session also exists (`taskWake.observeSession(...)` immediately before `runRealtimeSession(...)`), so the truncated patch samples were the cause of both questions.

**Issue 1 (sequence/pending consistency window) — real window, not reachable, and now closed anyway.** The only drainer of terminal updates is the foreground session: `DrainTerminalTasks` has a single caller in the session loop, and `runRealtimeSession` is called synchronously from the standby loop (no goroutine), so a drain cannot interleave with the wake case. The writer is the task worker, which appends, increments, and signals under one lock, so a consumed wake signal always implies the batch is already visible.

Even so, the reviewer's suggested fix is the right shape, so I applied it in `dde49405`: `Manager.TerminalState()` returns the pending updates and the sequence under a single `RLock`, replacing `PendingTerminalTasks`, and the standby wake case uses that one snapshot:

```go
pending, seq := agentTaskTerminalState(tasks)
if len(pending) == 0 { continue }
if !taskWake.shouldActivate(seq) { continue }
```

`go test -race -count=1 ./internal/agenttask/ ./cmd/daemon/` passes; the full `./cmd/... ./internal/...` suite only reports the two pre-existing PTY shell-tool failures that also fail on an unmodified baseline.

PR #661, opened 2026-09-09
feat(benchmark): remove use_procedure_steps test case

codesister:
This PR removes the `use_procedure_steps` benchmark case from `memory_v1` and updates the suite coverage test accordingly.

**Rationale**
- The prompt can cause the agent to enter the reimbursement workflow.
- Real execution is likely to time out, which makes the case unreliable as a memory benchmark requirement.
- This benchmark should avoid prompts that trigger long-running real-world workflows when the intended signal is memory recall.

**Changes**
- Remove the `use_procedure_steps` task block from `benchmark/suites/memory_v1.json`.
- Drop `use_procedure_steps` from the memory suite coverage assertion.
- Adjust the minimum expected memory suite task count.

**Validation**
- `PYTHONPATH=benchmark uv run pytest benchmark/tests/test_suite.py -k memory_suite_covers_representative_memory_behaviors`


aiden-lightning:

- PR: #661 — feat(benchmark): remove use_procedure_steps test case
- Head SHA: `5ed3aef830bfce927fbe9aba047f14825a0b8330`
- GitHub updatedAt: `2026-09-09T07:11:01Z`
- Reviewed at: `2026-09-09T15:22:18+08:00`

## Summary

The removal is clean and complete: `use_procedure_steps` appears in exactly two places in the repo (`benchmark/suites/memory_v1.json:225` and `benchmark/tests/test_suite.py:1252`), and the patch updates both. No runner code, docs, webui fixture, or `benchmark/scripts/quick_memory_test.py` reference the id, so nothing dangles.

One real (if minor) issue: both files lose their trailing newline.

## Verified Issues

### 1. Trailing newline stripped from both files — unrelated to the PR's purpose
**Confidence: HIGH**

- `benchmark/suites/memory_v1.json` — patch turns `}` into `}\ No newline at end of file`
- `benchmark/tests/test_suite.py:1738` — same on the closing `}` of the dict literal

Evidence: both files currently end with `0x0a` at the PR base (`tail -c 1 | od -c` → `\n` for each), as does `benchmark/suites/episode_memory_v1.json`. So the PR is actively removing an existing newline in two files whose only intended change is deleting a task.

Impact: no functional break — I checked `benchmark/pyproject.toml` (no ruff/black/flake8 config), `.pre-commit-config.yaml` (absent), `.coderabbit.yaml` (only golangci-lint, disabled), and `.github/workflows/ci.yml` (runs `python3 -m unittest tests.benchmark.test_docker_sandbox`, no pytest or Python lint over `benchmark/`), so nothing in CI will flag it. The cost is diff noise: the next edit near the end of either file will show a spurious two-line change, and `test_suite.py` now carries a W292 violation against the repo's own convention. Worth restoring both newlines.

## Observations

**The removed task looks genuinely unsatisfiable, which supports the change.** Its rubric (`steps_in_correct_order`, `memory_v1.json:235-237`) requires the agent to recall a procedure of 飞书填表 → approval → 招行 App 公务报销 → payment screenshot, but the task has no `setup` block, `global_reset` is `{}` (`memory_v1.json:5`), and grepping `招行`/`飞书填表`/`公务报销` across the repo hits only that task's own rubric text. The procedure it asks the agent to recall was never seeded anywhere. Note the neighboring `save_user_procedure` (`memory_v1.json:86`) stores a *different* procedure (date folder → invoice filename pattern → `expense.md`), so it wasn't the implicit source either.

**Coverage impact is limited.** Procedure *saving* is still covered by `save_user_procedure`, and procedure *application* is covered in the episode suites (`episode_reflection_v2.json:9` `verified_success_procedure_applies`, `episode_memory_after_v1.json:8` `qa_notes_title_save_procedure`).

**The count assertion is now inert.** `benchmark/tests/test_suite.py:1263-1264` asserts `expected_tasks <= set(task_by_id)` with 17 ids and then `len(suite.tasks) >= 16`. Since the suite has exactly 17 tasks after this change, the set check already implies ≥17, so the `>= 16` line can never fail on its own. This slack predates the PR (18 tasks with `>= 17`), so it's not a regression — but `>= 17` would have been the tighter update if the intent is to guard against silent removals.

**Same setup-less pattern elsewhere in the suite.** There are only 4 `setup` blocks in `memory_v1.json` (lines 145, 254, 311, 368), so `use_preference_language:173`, `use_rule_to_block_action:199`, `recall_session_chunk_details:282`, and `use_filtered_subset:339` also depend on state no setup seeds. Pre-existing and out of scope here, but if the reasoning behind this PR is "the rubric references memories that are never stored," those four are candidates for the same treatment or for added `seed_memory` setups.

## Verdict

Approve once the two trailing newlines are restored. The task removal itself is correct, complete, and well-justified by the missing seed data; there are no functional defects in the change.

aiden-lightning:

- PR: #661 — feat(benchmark): remove use_procedure_steps test case
- Head SHA: `24ec1df42fba94057be10cf40ca100169e266daf`
- GitHub updatedAt: `2026-09-11T08:12:44Z`
- Reviewed at: `2026-09-11T17:41:34+08:00`

I have everything I need. Two verified findings.

## Summary

The PR removes the `use_procedure_steps` task from `benchmark/suites/memory_v1.json`, dropping the suite from 18 to 17 tasks. The removal itself is clean (correct JSON object boundaries, no dangling commas, remaining structure intact), but it breaks a test that pins this task ID, and it strips the file's trailing newline.

## Verified Issues

### 1. Breaks `test_memory_suite_covers_representative_memory_behaviors` — HIGH confidence

`benchmark/tests/test_suite.py:1252` lists `"use_procedure_steps"` inside the `expected_tasks` set, which is enforced at line 1263:

```python
expected_tasks = {
    ...
    "use_rule_to_block_action",
    "use_procedure_steps",      # line 1252
    "recall_saved_fact_after_setup",
    ...
}
assert expected_tasks <= set(task_by_id)   # line 1263
assert len(suite.tasks) >= 17              # line 1264
```

`expected_tasks` holds 18 IDs. After this PR the suite contains 17 tasks, none of which is `use_procedure_steps`, so the subset assertion at line 1263 fails.

The `len(suite.tasks) >= 17` assertion at line 1264 still passes (17 >= 17), so it survives on the boundary — but it will break on the next task removal, and the test file is currently the only place recording the expected inventory.

Impact: the benchmark test suite fails. `benchmark/pyproject.toml:26-28` sets `testpaths = ["tests"]`, so `pytest` in the `benchmark/` directory collects and runs this test. The fix is to delete line 1252 and lower the count assertion at line 1264 to `>= 16` if you want to preserve the original one-task slack.

Note: I did not find a CI workflow that runs `benchmark/tests` — `.github/workflows/ci.yml:118` only runs `tests.benchmark.test_docker_sandbox`, a different path. So this failure likely surfaces locally rather than blocking the merge, which makes it easier to miss.

### 2. Trailing newline removed at end of file — MEDIUM confidence

The diff's final hunk shows `}` → `}` with `\ No newline at end of file`, an unrelated whitespace change. Every other suite JSON in `benchmark/suites/` ends with a newline (verified via `grep -c ""` line counts across all 18 files, which count the final line only when newline-terminated).

Impact: cosmetic, but it adds diff noise to the next edit of this file and diverges from the convention in the sibling suites. I found no CI lint enforcing end-of-file newlines, so this will not fail a build. Worth restoring since it is not related to the PR's purpose.

## Observations

The removal does not orphant anything else. `save_user_procedure` (`benchmark/suites/memory_v1.json:86-110`) still covers the save side of procedure memory, and its rubric checks `type == 'procedure'` plus step preservation, so procedure storage remains tested. What is lost is coverage of the recall-and-apply-in-order behavior — no other task in the suite exercises ordered procedure playback.

The removed task also had no `setup` block, unlike `use_preference_brevity` (line 145), `recall_saved_fact_after_setup` (line 254), and `forget_on_request` (line 368), which all seed state before running. Its `description_for_judge` assumed "a stored reimbursement procedure" and its rubric expected specific steps (飞书填表, 招行 App 公务报销) that appear nowhere else in the repo — grepping for those strings returns only the line being deleted. So the task depended on memory that nothing seeded, which is a plausible reason it was flaky and a reasonable basis for removal.

## Verdict

Request changes. The intent is sound and the JSON edit is correct, but `benchmark/tests/test_suite.py:1252` must be updated in the same PR or the memory suite test fails. The dropped trailing newline is worth reverting as unrelated noise.

HuangHan96 (approved):

PR #667, opened 2026-09-10
fix(realtime): check outstanding tasks before creating background work

HuangHan96:
## Problem

The foreground realtime agent could start the same device work twice. A second `create_agent_task` for work that was already in flight was accepted like any other call: it returned a brand-new `task_<uuid>` with `status: queued`, and the background agent then ran it again once the first task finished.

This is visible in a device log (`agent.log`, 2026-09-09) from one realtime chat (`benchmark-6b2f…`, 03:40:25–03:40:35):

| time | event |
| --- | --- |
| 03:40:27 | `create_agent_task` → background run starts: `处理差旅报销流程：用户本次出差打车费用为800元…` |
| 03:40:28 | `query_agent_task` (no result yet) |
| 03:40:29 | `create_agent_task` again, same request period |
| 03:41:32 | first run finishes and writes the expense record; the queued duplicate starts immediately: `重新处理：用户报销本次出差打车费用800元…` |

The second run redid the same 800 CNY reimbursement, and by then the phone had already moved on to another task, so it started searching the home screen for a reimbursement app.

`query_agent_task` could not have prevented this: `task_id` was required, so the model had no way to see what it had already started. The prompt also said nothing about checking first.

## Change

**`internal/agenttask`** — new `Manager.Outstanding()`: tasks that have not finished, plus finished tasks whose result has not been delivered to the foreground yet, newest first. Delivered results are deliberately excluded, because the foreground holds them in its own context, so repeating the same request later is a genuine new request rather than a duplicate.

**`query_agent_task`** — `task_id` is now an optional filter. Omitted → every outstanding task; given → the same envelope with that task in it. Both modes answer `{"tasks": [...]}`, an empty list encodes as `[]` (not `null`), and an unknown `task_id` stays `{"error":"agent task not found"}` so an empty list is never misread as "nothing in flight". `cancel_agent_task` still requires an id.

**Prompt** — `DefaultRealtimeVoiceInstructions` and the `create_agent_task` description both require `query_agent_task` with no `task_id` before creating work, and continue the existing task when it already covers the request.

Semantic judgement is the point here: the duplicate in the log was re-worded (`处理差旅报销流程…` vs `重新处理…`), so exact-prompt dedup in the manager would not catch it. The foreground just needs to see what is in flight.

## Verification

- `go build ./...` and `go test ./...` in `src/agent`.
- New tests: `TestManagerOutstandingTracksInFlightAndUndeliveredWork` (in-flight work is outstanding; a finished result stays visible until delivered, then drops out), `TestRealtimeQueryTaskWithoutTaskIDListsOutstandingWork` (empty list encoded as `[]`, unknown id is an error), and `TestRealtimeTaskCreationRequiresOutstandingWorkCheck`, which keeps the prompt rule and the tool schema in step — the rule only works while `task_id` is optional.
- Full `go test ./...` failure set is byte-identical to a clean checkout of `main` (16 pre-existing environment-dependent failures: unix sockets, adb, special files); no new failures. `gofmt` clean.
- Docs: `docs/04-agent/foreground-background-agents.md` tool table and task-creation section.

## Tradeoff

This is a prompt-level constraint, not an enforced one: the model can still skip the check. The extra `query_agent_task` round trip costs about a second per creation. Hard enforcement (rejecting a create whose prompt exactly matches a non-terminal task) is possible but would not have caught the case above, so it is left out.

## Summary by CodeRabbit

* **New Features**
  * Added support for viewing all outstanding background agent tasks when no task ID is provided.
  * Task results now use a consistent list format, including when no tasks are available.
  * Background task creation now checks for existing in-progress or undelivered work to help prevent duplicates.

* **Documentation**
  * Updated the foreground tool catalog to describe task-list queries and duplicate-task handling.


aiden-lightning:

- PR: #667 — fix(realtime): check outstanding tasks before creating background work
- Head SHA: `8b43233a0963d74a08a2b7d1afd05aabaf195fb7`
- GitHub updatedAt: `2026-09-10T09:56:46Z`
- Reviewed at: `2026-09-10T17:58:29+08:00`

Based on my review of PR #667, I found one **verified defect** in the code:

## Summary

The PR adds functionality to check for outstanding background tasks before creating duplicates in the realtime voice system. The implementation is mostly sound, but there's a **critical compilation error** due to a missing import.

## Verified Issues

### CRITICAL: Missing `slices` import causes compilation failure

**File**: `src/agent/internal/agenttask/manager.go:6`

**Evidence**: The PR diff shows line 6 in manager.go adds `"slices"` to the imports, but examining the actual file in the repository shows the imports section only contains:
```go
import (
    "context"
    "errors"
    "strings"
    "sync"
    "time"
    
    "github.com/google/uuid"
)
```

The `slices` package is missing, but line 211 in the Outstanding() method uses `slices.SortFunc()`:
```go
slices.SortFunc(tasks, func(a, b Task) int {
```

**Impact**: The code will fail to compile with an "undefined: slices" error. This breaks the build completely.

**Confidence**: HIGH - Verified by reading the actual file content and cross-referencing with the diff.

### MEDIUM: Test helper `decodeRealtimeTaskQuery` appears to be missing

**File**: `src/agent/cmd/daemon/main_test.go:330`

**Evidence**: The PR diff shows the test `TestRealtimeAgentTaskToolsAreNonBlocking` was modified to call `decodeRealtimeTaskQuery()`:
```go
running := decodeRealtimeTaskQuery(t, executor.call(...))
```

However, searching the test file shows no definition of this helper function. The current repository version at line 330-333 shows the old code that directly unmarshals to a single Task, not using this helper.

**Impact**: If the diff is applied as shown, tests will fail to compile due to the undefined helper function. This suggests either:
1. The helper function definition is missing from the patch
2. The patch wasn't properly generated
3. The function should be defined but isn't shown in the truncated diff

**Confidence**: MEDIUM - The diff clearly references this function but I cannot verify if it exists elsewhere in the file since the patch appears truncated.

## Observations

The core logic looks correct:
- The `Outstanding()` method properly filters tasks to show in-flight + undelivered terminal tasks
- The `terminal()` helper correctly identifies terminal statuses
- The tool schema change making `task_id` optional is implemented correctly
- The instructions and documentation properly explain the duplicate-prevention pattern
- Test coverage for the new Outstanding() method is comprehensive

## Verdict

**BLOCK** - The missing `slices` import is a critical build-breaking issue that must be fixed before merge. The potentially missing test helper should also be verified.

PR #669, opened 2026-09-11
fix(agent): keep finished shell sessions pollable

HuangHan96:
## Problem

Polling an already-exited, fully drained background shell session **deleted** it, so the next poll answered:

```
shell session not found: shell-1
```

That string is indistinguishable from a bogus `session_id`, and it is returned with `CodeInvalidArguments` — so a model doing a redundant confirm-poll is told its *parameters* were wrong, and is invited to start a replacement session (which the prompt explicitly warns against).

Device log evidence, same session, 34s apart:

| # | call | result |
|---|---|---|
| `[293]` | `shell action=poll session_id=shell-1` | `{"exit_code":0,"output":"23 24 25\nsum: 72\n","running":false,…}` |
| `[295]` | `shell action=poll session_id=shell-1` | `shell session not found: shell-1` |

The first poll both delivered the output **and** destroyed the session, because the process had already exited and the poll drained the ring buffer:

```go
if !running && !session.output.hasUnread() {
    globalShellSessionManager.delete(session.id)
}
```

## Change

- Draining the last bytes now calls `session.output.release()` instead of deleting the session, so a repeated poll answers with the terminal payload. Release is only reachable once every produced byte has been read (the existing `!hasUnread()` guard means exactly that), so no output is lost.
- `start` and `poll` now report `completed`; `poll` adds `finished_at` once the process exits. "This will produce nothing more" is explicit instead of implied by `running=false`.
- Eviction is unchanged — the record is still reclaimed on the 30-minute idle timeout.
- `stop()` no longer signals a process that `Wait()` has already reaped. Retaining finished sessions means they now reach eviction, which calls `stop()`, and the reaped PID may since have been reused.

Retention drops the ring buffer (200 KB, eagerly allocated) and keeps only session metadata, so it does not hold output bytes.

## Verification

- `gofmt`, `go build ./...`, `go vet ./internal/agent/` — clean
- `go test -race -count=1 -run TestShell ./internal/agent/` — pass
- `TestShellToolPollAfterCompletionIsIdempotent` reproduces the log sequence: the first poll yields `sum: 72`, the repeat poll returns the terminal state instead of an error
- `TestShellToolBackgroundPollDrainsAndKeepsSession` replaces `TestShellToolBackgroundPollDrainsBeforeDelete`, which asserted the old drop-after-drain behaviour

No on-device or live-model validation performed.

## Not in this PR

`session not found` is still a bare string carrying `CodeInvalidArguments`. With sessions retained, that path now only fires for ids that never existed or have expired, so the misleading case is gone. Giving bare-string tool failures a structured channel is a follow-up that also covers the `web_search` backend-failure case (a rate-limited search is currently indistinguishable from "no results") and artifact re-reads being re-artifactized.

## Summary by CodeRabbit

- **New Features**
  - Background shell sessions now report completion status and, when available, the completion time.
  - Completed sessions remain available with stable identifiers for follow-up polling.

- **Bug Fixes**
  - Completed sessions no longer disappear immediately after their output is consumed.
  - Repeated polls provide consistent completion details without repeating previously retrieved output.
  - Stopping an already-completed session no longer risks affecting another process.

### DIFF
--- a/src/agent/internal/agent/tools_shell.go
+++ b/src/agent/internal/agent/tools_shell.go
@@ -206,6 +206,7 @@ func shellStartBackground(ctx context.Context, arguments map[string]interface{},
 	return shellJSONString(map[string]interface{}{
 		"session_id": session.id,
 		"running":    true,
+		"completed":  false,
 		"pid":        session.processID(),
 		"command":    session.command,
 		"workdir":    session.workdir,
@@ -238,6 +239,7 @@ func shellPollBackground(ctx context.Context, arguments map[string]interface{}) 
 	payload := map[string]interface{}{
 		"session_id":       session.id,
 		"running":          running,
+		"completed":        !running,
 		"pid":              session.processID(),
 		"command":          session.command,
 		"workdir":          session.workdir,
@@ -248,9 +250,16 @@ func shellPollBackground(ctx context.Context, arguments map[string]interface{}) 
 		"exit_code":        session.exitCodeValue(),
 		"exit_error":       session.exitErrorText(),
 	}
+	if finished := session.finishedAtOr(); !finished.IsZero() {
+		payload["finished_at"] = finished.Format(time.RFC3339)
+	}
 
 	if !running && !session.output.hasUnread() {
-		globalShellSessionManager.delete(session.id)
+		// Keep the session so a repeated poll answers with this terminal payload
+		// instead of "session not found". Drop the bytes instead: every produced
+		// byte has been delivered, so nothing is lost, and the record stays small
+		// until evictIdle reclaims it on a fixed clock.
+		session.output.release()
 	}
 	return shellJSONString(payload), nil
 }
--- a/src/agent/internal/agent/tools_shell_session.go
+++ b/src/agent/internal/agent/tools_shell_session.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -41,6 +42,7 @@ type shellSession struct {
 
 	mu     sync.Mutex
 	closed bool
+	reaped bool
 }
 
 type shellRingBuffer struct {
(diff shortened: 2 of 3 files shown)


aiden-lightning:

- PR: #669 — fix(agent): keep finished shell sessions pollable
- Head SHA: `3c9e76fb14a2b0cf26c84048200e0bf1ef0a51dc`
- GitHub updatedAt: `2026-09-11T06:07:12Z`
- Reviewed at: `2026-09-11T14:14:24+08:00`

Based on my thorough review of PR #669, I've analyzed the code changes, verified the logic, and checked for potential issues. Here's my assessment:

## Summary

This PR fixes a usability issue where polling a finished shell session would return "session not found" after output was drained, making it indistinguishable from an invalid session ID. The fix keeps finished sessions pollable by retaining the session record while releasing buffered output, and adds a critical safety fix to prevent signaling potentially reused PIDs.

## Verified Issues

**None found.** The implementation is sound.

## Observations

### Changes Are Well-Designed

1. **Memory Management**: The `release()` method correctly clears the buffer while keeping the session record. Subsequent polls safely handle released buffers by returning empty output.

2. **Idempotent Operations**: Calling `release()` multiple times is safe (sets `buf=nil`, `readPos=0`, `truncated=false` repeatedly). The condition `!running && !hasUnread()` ensures release only happens when no new output can arrive.

3. **Critical Safety Fix in `stop()`**: The new check preventing signals to finished sessions is important. Without it, the code could signal a reused PID:
   - After `Wait()` returns, the OS may reuse the PID
   - The added `!s.isRunning()` check prevents signaling a potentially different process
   - This is verified using the done channel, which closes only after `Wait()` completes

4. **Proper Cleanup**: Sessions are still evicted after 30 minutes of inactivity via `evictIdle()`, preventing unbounded memory growth.

5. **Thread Safety**: All buffer operations properly use mutexes. The check-then-release sequence in `shellPollBackground` is safe because `!running` guarantees no new output.

6. **Test Coverage**: Tests verify the new behavior including idempotent polling, terminal state reporting, and the `completed`/`finished_at` fields.

## Verdict

**LGTM** ✓ 

The changes are isolated, well-tested, and fix both a usability issue (idempotent polling) and a correctness issue (avoiding signals to reused PIDs). The implementation properly handles edge cases including concurrent access, multiple releases, and cleanup timing.

PR #690, opened 2026-09-16
fix(realtime): enforce configured response language

ljunb:
## Summary
- append the shared configured-locale guidance to every realtime session
- remove the conflicting default instruction to follow the detected input language
- document that response locale does not constrain provider-native transcription

## Root cause
The regular agent prompt enforced the configured locale, but realtime sessions only sent voice_model.instructions. Gemini Live could therefore carry a language inferred from audio or prior turns into later responses, including replying in Japanese to English input.

## Tests
- go test ./cmd/daemon ./internal/agent/realtimevoice
- go test ./cmd/daemon -run TestRealtimeSessionConfig -count=1
- relevant packages in go test ./... passed; unrelated screenprovider and OTA tests require host Unix socket/tooling support unavailable in this environment

## Summary by CodeRabbit

* **New Features**
  * Realtime voice responses now follow the configured agent locale, including locales such as Chinese (China) and English (United States).
  * Response-language guidance is consistently applied across agent entry points and custom voice instructions.

* **Documentation**
  * Clarified that the agent locale controls response language but does not configure or restrict input transcription language.


HuangHan96 (approved):

aiden-lightning:

- PR: #690 — fix(realtime): enforce configured response language
- Head SHA: `33a72a2eb51b583f342ce8d33cc369a1c92fc2d1`
- GitHub updatedAt: `2026-09-16T06:34:17Z`
- Reviewed at: `2026-09-16T14:55:51+08:00`

Verified the patch against the repository code. I could not run `go test`/`go vet` (command approval was declined), so compile/test claims below are from reading code, not execution.

## Summary

The change appends the shared response-language rule to the realtime session instructions and reuses `responseLanguageGuidance` via a new exported wrapper. Wiring is correct and narrowly scoped:

- `strings` is already imported in both `src/agent/cmd/daemon/realtime_wakeup.go:11` and `src/agent/cmd/daemon/main_test.go:13`, so the new `strings.*` uses compile.
- `realtimevoice.SessionConfig` is constructed in exactly one non-test place (`realtime_wakeup.go:472`), and nothing writes back to `cfg.VoiceModel.Instructions` (no matches for `VoiceModel.Instructions =`), so the guidance cannot be appended twice or accumulate across reconnects.
- Both pre-existing exact-equality assertions on `got.Instructions` (`main_test.go:55`, `main_test.go:112`) are the only ones in the tree and both are updated to `HasPrefix`, so no other test is silently broken.
- The new test's expectations match the implementation: `responseLanguageGuidance` (`prompt.go:96-108`) emits `"The configured response locale is <locale>."` and `"Always respond in <language>"`, and `normalizeResponseLocale` (`prompt.go:85-94`) maps `zh-CN`→Simplified Chinese, `en-US`→English. `Config.Validate` (`config.go:1468-1473`) restricts locale to those two values, so the table covers the full domain.
- The doc claim is accurate: no realtime provider payload sets a transcription language field (`grep -i language` over `internal/agent/realtimevoice/*.go` only matches Gemini endpoint hostnames), and OpenAI only pins `DefaultOpenAIInputTranscriptionModel` (`openai.go:18`).

## Verified Issues

### 1. `<tts>` guidance is injected into realtime instructions, but the realtime path has no `<tts>` handling — MEDIUM

`src/agent/internal/agent/prompt.go:104` (text now reused by `realtime_wakeup.go:611`)

The shared rule includes:

> Write all conversational assistant content in `<language>` from the first generated token. This includes progress messages, final answers, and **all text inside `<tts>...</tts>`**.

That sentence was written for the text-agent path, where `<tts>` is a hard requirement (`defaultAgentBehavior`, `prompt.go:148`) and output is unwrapped by `speech.BuildText`/`speech.ExtractText` before display. The realtime path has neither half of that contract:

- `agent.DefaultRealtimeVoiceInstructions` (`config_defaults.go:16-27`) never mentions `<tts>`, so the appended text is the only place the realtime model hears about the tag.
- There is no tag stripping on the realtime output path. `grep` for `<tts>`/`ExtractTTS`/`speech.` finds nothing in `internal/agent/realtimevoice/` or `realtime_wakeup.go`; the only `tts` imports there are the audio-format/resampler package (`realtime_wakeup.go:20`, `managed.go:10`). The four `speech.BuildText`/`ExtractText` callers are all non-realtime (`server.go:2289`, `audio_dialog.go:1070`, `runtime.go:237,244`, `live_activity.go:499,506`).

Failure scenario: with `input_mode` realtime and a text-capable provider, the model follows the mention of `<tts>` and emits a literal wrapper. `event.Text` accumulates into `responseText`/`chatText` and is forwarded verbatim as a chat delta (`realtime_wakeup.go:1471-1474`) and persisted verbatim (`appendRealtimeAssistantMessage`, `realtime_wakeup.go:2047-2052`) — so `<tts>` shows up in the config-web chat transcript and in stored history, and since realtime audio is synthesized from the model's own text, the tag can also be spoken. Dropping that clause (or passing a realtime-specific variant of the guidance) avoids introducing a convention the realtime path cannot honor.

## Observations

**Changing locale now resets the realtime conversation session.** `runRealtimeSessionWithRegistry` passes `sessionConfig.Instructions` as the context-manager system prompt (`realtime_wakeup.go:863`), and `InitializeContextManager` starts a brand-new session whenever the stored system prompt differs (`context_manager_bridge.go:211-217`, compared in `contextManagerHasSystemPrompt:230-237`). Because instructions now depend on locale, switching `agent.locale` — and the first run after this upgrade — discards the current realtime session instead of resuming it. That matches the documented intent of that branch ("configuration changes require a different system prompt") and older turns remain reachable via `recall_session_chunks`, so this looks acceptable; flagging it since it's a new trigger for session rotation.

**The guidance is appended unconditionally, including over custom instructions.** `realtime_wakeup.go:611` runs after the empty-string fallback, so a user who sets `[voice_model] instructions` with their own language rule gets the configured-locale rule appended after it. Given the PR title that appears deliberate, and the appended text is last so it wins — just noting the operator-visible behavior change.

**`ResponseLanguageGuidance` is a pure pass-through, and `LocaleOrDefault()` at the call site is redundant.** `responseLanguageGuidance` already calls `normalizeResponseLocale`, which falls back to `defaultLocale` for empty/unknown input (`prompt.go:85-94`), so `cfg.LocaleOrDefault()` (`config.go:1822-1828`) cannot change the result. Harmless; the existing caller passes raw `cfg.Locale` (`role_profile.go:61`). Exporting the original function instead of adding a wrapper would avoid the near-duplicate name pair, but that's a style call.

## Verdict

One MEDIUM-confidence issue worth addressing: the reused guidance carries a `<tts>` clause into a path with no `<tts>` parsing or stripping, which can surface literal tags in realtime chat text, persisted history, and speech. Everything else checks out — single construction site, no double-append, tests updated consistently with the implementation and covering both valid locales, and the doc note matches the provider payloads.

PR #698, opened 2026-09-17
fix(config-web): align the recovery verdict across surfaces

ZhaoQing419:
## Summary

Follow-ups on the audit of #695. This branch is stacked on that PR's head
(`271a8f80a`), so the diff against `main` contains #695 plus one commit. Land #695
first, or cherry-pick `bbd529002`.

Three places where two surfaces still disagreed, or a supported state still
misread.

### 1. An absent `agent.toml` was reported as an invalid configuration

The opt-in rule for optional speech providers is keyed off what the file
declared, and `loadResolvedConfig()` ran it against a file that is not there.
`metadata` is zero for a missing file, so it read as "declared nothing" and
cleared the built-in defaults:

| `agent.toml` absent | before #695 | after #695 | with this commit |
| --- | --- | --- | --- |
| `tts.provider` / `stt.provider` in the editor view | `minimax-cn` / `openai-whisper` | `""` / `""` | `minimax-cn` / `openai-whisper` |
| `config_valid` | `true` | `false` | `true` |
| details | empty | `decode TOML config: open ...: no such file or directory` | empty |

Shipped images always contain the file (`container-assemble-images.sh` refuses to
stage an image without it, `container-audit-images.sh` verifies it, OTA preserves
`userdata`), so this is a dev/recovery path rather than a device regression — but
it is the documented first-boot behavior of `loadResolvedConfig`, and reporting a
device that has never been configured as invalid is the wrong answer for the page
that exists to create the file. `applyRuntimeOptionalProviderDefaults` now takes
whether the file was read at all.

### 2. `config-check --config` and the portal reported different verdicts

The portal read the runtime loader; the CLI read the strict editor view. The live
case is a persisted `input_mode = "realtime"` whose credential was removed: the
Agent boots by falling back to text mode and the portal reports the config as
valid, while `config-check` reported

```
voice_model.api_key is required when input_mode=realtime
```

for a field the runtime does not require. Both now read one
`agent.ConfigValidationVerdict`, so the verdict shown and the verdict that gates
startup cannot drift. Candidate values a user is about to save are still validated
strictly by `config-check --stdin`, and the difference is documented.

### 3. The live STT test refused a recoverable config

`POST /api/config-test/stt/start` loaded its context through the strict resolver,
so it answered `503 load Agent config: stt.provider is required when input_mode=stt`
in the recovery state that makes it worth pressing — the flagged `stt.provider` is
what the user is testing a replacement for. It now reads the persisted context
without semantic validation, matching the section Test buttons fixed in #695, while
a file that cannot be read or decoded still fails.

Also dropped the now-unused `parseValidationErrors` indirection in `checkConfig`,
and refreshed the two comments that explained the legacy voice migration gate in
terms of loaders that no longer differ.

## Verification

- `go test -count=1 ./internal/configweb/... ./internal/configupdate/... ./cmd/daemon/...`
- `go test -count=1 ./internal/agent/` (two pre-existing PTY failures also fail on `b922208`)
- `node --experimental-vm-modules tests/config_web_{api,dynamic_form,i18n,storage}_test.mjs`
- each of the three fixes has a test that fails when the fix is reverted, checked by
  reverting it locally


ZhaoQing419:
Superseded: the same three fixes now live on this PR's branch as commit 40982b05c, so #695 is the single PR for the recovery work. Closing this stacked branch.

PR #699, opened 2026-09-17
feat(agent): add Google Gemini provider on native Interactions API

ZhaoQing419:
Register a \`gemini\` model provider that speaks Gemini's native Interactions API at \`https://generativelanguage.googleapis.com/v1beta/interactions\` and authenticates with \`x-goog-api-key\`.

## Why native Interactions instead of OpenAI-compatible path

The OpenAI-compatible \`/chat/completions\` endpoint cannot round-trip the \`thought_signature\` that Gemini 3 models require on replayed function calls, so multi-turn tool use fails there with \`400 Function call is missing a thought_signature\`. Google also documents \`generateContent\` as legacy and recommends the native API.

\`api_mode\` therefore accepts only:
- \`interactions\` (local transcript, \`store=false\`)
- \`interactions_stateful\` (stored interaction chained by \`previous_interaction_id\`)

An unset \`api_mode\` selects \`interactions\`, and OpenAI-compatible modes are rejected at config validation.

## Transcript fidelity

Stored Interactions steps are replayed byte-for-byte, keeping the thought signatures the provider validates:
- Turns that predate this transport carry their tool exchanges as text instead (unsigned replayed calls are rejected outright)
- Compaction and session revisions keep \`InteractionsSteps\` — unlike a Responses ID they're not a server-side state pointer but the steps replayed in the next request
- Streamed function-call arguments replace the \`step.start\` placeholder rather than extend it (the API sends \`arguments\` twice in different shapes: as an object on \`step.start\`, as JSON text on each \`arguments_delta\`; appending both produced \`{}{"command":"date"}\`)

## Config Web improvements

- \`ExcludeProviders\` mechanism hides the empty \`api_mode\` option ("the provider's compatible default") for providers with no compatible transport
- Switching provider now re-filters the \`api_mode\` options instead of leaving a stale Interactions choice on screen

## Documentation

Also backfills \`deepseek\` into \`docs/03-services/agent.md\` and \`docs/04-agent/overview.md\`, which PR #674 missed.

## Verification

Wire format verified against real device traffic: \`thought_signature\` arrives top-level in \`step.delta\`, and all five request step types (\`user_input\`, \`model_output\`, \`thought\`, \`function_call\`, \`function_result\`) are accepted.

All 13 Interactions tests pass, including a new regression test for the streamed arguments placeholder issue.

## Summary by CodeRabbit

- **New Features**
  - Added native Google Gemini support through the Interactions API.
  - Added stateless and stateful conversations, streaming, tool use, multimodal content, and context replay.
  - Added Gemini model listings, including a recommended model, to configuration and model selection.
  - Added provider-specific API modes and temperature defaults.

- **Bug Fixes**
  - Improved replay of Gemini thought signatures and tool-call metadata.
  - Added handling for expired interaction state and unsupported historical tool exchanges.

- **Documentation**
  - Documented Gemini setup, supported models, reasoning levels, API keys, endpoints, and configuration guidance.


aiden-lightning:

- PR: #699 — feat(agent): add Google Gemini provider on native Interactions API
- Head SHA: `28962354527a6c2634d8e76a9340624ec85ab851`
- GitHub updatedAt: `2026-09-17T09:42:38Z`
- Reviewed at: `2026-09-18T10:32:24+08:00`

## Summary

Reviewed the highest-risk files in this PR: the new Gemini Interactions API implementation (`interactions_model.go`, 1172 lines), its integration into the agent loop, config validation, and the provider registry. Found one confirmed defect in the agent loop integration.

## Coverage

**Files reviewed (8 of 30):**
- `src/agent/internal/agent/interactions_model.go` (new file, 1172 lines)
- `src/agent/internal/agent/interactions_model_test.go` (new file, 611 lines)
- `src/agent/internal/agent/agent_loop.go` (integration points)
- `src/agent/internal/agent/model_provider_registry.go` (provider registration)
- `src/agent/internal/agent/config.go` (validation logic)
- `src/agent/internal/agent/messages/messages.go` (message structure)
- `src/agent/internal/agent/compactor/compactor.go` (compaction integration)
- `src/agent/internal/agent/models_test.go` (builder tests)

**Files not reviewed (22):** All remaining changed files including docs, config web UI, model specs, and additional test files. These were skipped to stay within the time budget and focus on the core implementation.

## Verified Issues

### 1. Logic error in agent loop step filtering

**File:** `src/agent/internal/agent/agent_loop.go:891`

**Issue:** The condition `metadata.Type != "function_call" || strings.TrimSpace(metadata.ID) == toolID` keeps a function_call step when EITHER it's not a function_call OR its ID matches. This is backwards. When `metadata.Type != "function_call"` is true (for thought, model_output, etc.), the step is kept regardless of the `||` right-hand side. When it IS a function_call, the step is kept only if the ID matches the executed tool. But the goal stated in the comment is to "remove unexecuted native Gemini function-call steps" — meaning non-matching function_call steps should be filtered out, while all other step types should be kept.

**Evidence from code:**
```go
// Line 891
if json.Unmarshal(item, &metadata) != nil || metadata.Type != "function_call" || strings.TrimSpace(metadata.ID) == toolID {
    filtered = append(filtered, item)
}
```

The logic says: keep the item if (unmarshal fails) OR (not a function_call) OR (ID matches). The second condition `metadata.Type != "function_call"` should be `metadata.Type == "function_call"` to properly filter only function_call steps, or the entire clause needs restructuring.

**Correct logic should be:** Keep all non-function_call steps, and among function_call steps, keep only the one matching toolID.

**Failure scenario:** When the Interactions response contains multiple function_call steps and the agent executes only one, the subsequent request will replay all the unexecuted function_call steps without corresponding function_result steps. Gemini will reject this with HTTP 400 because every function_call must have a paired function_result before the next request.

**Impact:** Multi-tool responses will fail on the next agent loop iteration when using Gemini in `interactions` mode.

## Verdict

**One confirmed defect** in the agent loop integration logic that will break multi-tool interactions. The core Interactions implementation appears structurally sound — it handles streaming, step reconstruction, thought signature preservation, and stateful/stateless modes correctly. Config validation properly enforces the new API modes and provider capabilities.

ZhaoQing419:
Addressed the current CI/review feedback in cfd61fcd. The config metadata validator now accepts providerType, model.api_mode hydration drops values unsupported by the newly selected provider, and the temperature guidance now describes Agent-selected versus provider-managed defaults. I also reviewed the automated agent_loop.go:891 finding: the existing predicate is correct because it keeps every non-function step and only keeps a function_call whose ID matches the executed tool. TestChoiceWithOnlyToolCallDropsUnexecutedInteractionsSteps now covers thought/model_output preservation, unmatched-call removal, matched-call retention, and clone-on-filter behavior.

HuangHan96 (approved):

PR #701, opened 2026-09-18
fix(config-web): save voice mode dependencies atomically

ljunb:
## Summary
- save voice mode and its active provider references in one config update
- include STT and TTS settings when switching to Classic mode
- automatically select newly created model, STT, TTS, and realtime provider records
- preserve unsaved selections and editable state when validation fails

## Root cause
Config Web persisted `input_mode` separately from the selected realtime provider. Each PATCH is fully validated by the backend, so switching to `realtime` could be rejected before `voice_model.provider` was saved. Newly created provider records also were not selected, which encouraged duplicate records such as `alicloud-2`.

## Tests
- all `tests/config_web_*_test.mjs`
- `go test ./internal/configupdate ./cmd/daemon`
- JavaScript syntax checks
- `gofmt -d src/agent/internal/configupdate/update_test.go`
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Added unified saving for voice-mode settings, including realtime voice model, speech-to-text, and text-to-speech options.
  - New voice providers are automatically selected after creation, reducing setup steps.
  - Provider and voice settings now save together, helping prevent partially applied configuration changes.

- **Bug Fixes**
  - Improved configuration reliability when switching voice modes or adding providers.
  - Prevented unrelated voice settings from being changed when saving a new provider.
  - Configuration errors are now highlighted on affected fields and cleared as corrections are made.
  - Added clearer recovery messaging when an invalid configuration is loaded.


ljunb:
Delta link: https://delta.dev/join/thread_01m2qce43hcevstqb0mcs7c6pa#kgKTkgCSAcQBAJKSks4AAmD4kgHEELst3LkRikz3pZXz-bYspCzCAAA

HuangHan96 (approved):

PR #703, opened 2026-09-18
test(benchmark): validate MobileGym scroll behavior and timing

ZhaoQing419:
The MobileGym scroll benchmark needs both velocity-dependent fling and preservation of agent-selected swipe timing. The MNK adapter previously dropped `duration_ms`, so identical paths requested at 120 ms and 2400 ms both became 160 ms swipes.

This PR adds deterministic environment-state and route assertions, explicit foreground-app setup, a two-case Scroll Lab regression suite for positions 24 and 83, an optional 100-depth calibration sweep, and a fixed-distance velocity probe. The monotonic `maxFirstVisibleOrdinal` high-water mark records overshoot even when the agent later scrolls back. Timed-out bridge work is cancelled to prevent late setup mutations.

The duration fix from #709 is included here: swipe and drag preserve positive `duration_ms`, validate the integer range `0..10000`, and retain backend defaults for omitted or zero durations. HTTP-to-environment tests cover this path.

Dependency: [AidenAI-IO/mobilegym#3](https://github.com/AidenAI-IO/mobilegym/pull/3) is merged. The submodule now points to its main-branch commit `d0549195beccbe08f963bf7758bef3a65fe3ac08`, replacing the PR-head commit `79cb44b629b55f3a60d859b47e332fe9c1bbe000`. Both commits have the same source tree. This addresses the requested submodule update.

Validation on the combined branch:

- 493 focused tests passed across MNK, MobileGym, environment assertions, setup/recovery, task evaluation, suite parsing, and the shared ADB, vphone, desktop, and mock bridges.
- One previously documented, unrelated baseline test was deselected: `test_memory_suite_covers_representative_memory_behaviors` still expects `use_procedure_steps`, which is absent from `memory_v1.json`.
- Initialized the MobileGym submodule from its configured AidenAI-IO URL and verified the pinned commit is advertised as `refs/heads/main`.
- `git diff --check` passed.

The earlier live simulator validation recorded in #709 showed that every requested duration reached the environment unchanged, with shorter swipes producing greater fling distance. This update reruns the automated tests; it does not claim a new real-device calibration or full LLM benchmark run.

Task context: UI scroll optimization research (`d979c419-cf57-41f5-8b92-b26f1ffe17eb`), currently in the design section. No research findings or experiment artifacts are included.


HuangHan96 (changes requested):
需要更新 Submodule mobilegym

HuangHan96 (approved):

PR #713, opened 2026-09-20
fix(config): apply model changes without draining voice sessions

ZhaoQing419:
Changing Model Settings could leave configuration pending for the duration of an active STT or realtime voice conversation. The voice lifecycle compared the ordinary `Model` configuration and waited for the voice loop to exit before Runtime could replace the model client.

Ordinary model/provider changes now wait only for the active Agent task. Subsequent tasks use the replacement client while the voice session stays open. Voice component changes retain their existing drain behavior. Run-start logs report the active provider/model instead of the dialog's startup snapshot, and the configuration policy documents the boundary.

Connection warming reads the applied ordinary model endpoint at each recording start while retaining the active dialog's STT endpoint. Adding, replacing, or clearing the model endpoint takes effect without rebuilding the voice dialog; in-flight warmups retain their own endpoint snapshot.

Validation:
- Warmup regression tests cover endpoint replacement, enabling an initially absent endpoint, clearing it while retaining STT warming, and updates during an in-flight warmup. All three dialog cases fail before the endpoint-refresh fix and pass afterward.
- Added a regression covering STT and realtime modes with a held voice session and two HTTP provider fixtures. It checks that an in-flight task completes on the old provider, the next task reaches the new endpoint with the new model and credentials, and the voice loop stays open. The test fails before the fix and passes afterward, both on macOS and on the ARM board.
- Ran `go test ./...` with optional live-provider tests disabled. All packages passed across that run and targeted reruns: `screenprovider` and `configweb` initially hit macOS Unix-socket path-length, timeout, or temporary-directory cleanup failures, then both passed with `TMPDIR=/tmp go test -p 1` without code changes.
- Focused race tests passed for recording warmups, concurrent endpoint updates, input lifecycle, model replacement, and configuration application.
- Cross-compiled the final Agent with Go 1.26.0 for Linux ARMv7. The endpoint-refresh addition has not been deployed to hardware. For model-lifecycle hardware testing, applied the same patch to the board's installed `a70bffcc2617` source revision, preserving its text-mode configuration; the compatible build passed lifecycle tests on the board and was deployed with both Agent and Config Web healthy.
- Live Config Web switching on the board: DeepSeek completed a chat; changing to OpenRouter sent the next request as `google/gemini-3.5-flash` with the same Agent PID/runtime ID. OpenRouter rejected it with HTTP 402 for insufficient credits, so successful completion there was not verified. Restored the original TOML byte-for-byte and confirmed it applied. The active-voice regression uses a controlled voice loop and local provider fixtures because this board has no configured realtime voice credentials.

## Summary by CodeRabbit

* **New Features**
  * Model and provider configuration updates now apply after the current agent task finishes.
  * Active voice sessions continue uninterrupted when model settings change.
  * Subsequent agent tasks use the updated model configuration.
  * Connection warmups now refresh automatically when model endpoints change.

* **Bug Fixes**
  * Voice-session reload detection no longer restarts or drains the voice loop for model-only changes.
  * Updated endpoint configurations are applied reliably during connection warmups.

* **Documentation**
  * Updated configuration guidance to describe the separate handling of model and voice settings.


HuangHan96:
@coderabbitai review

HuangHan96 (approved):

